Issue #42, opened 2013-10-18
jsonld.flatten README example inconsistent with implementation


bollwyvl:
If `jsonld.flatten` is called as suggested in the README:

``` javascript
jsonld.flatten(doc, function(err, flattened) {
  // all deep-level trees flattened to the top-level
});
```

the callback will never be called. It looks like the second argument to flatten **must** be a context object or `null`.

Here are two options I see:
### Change README

Calling with a blank context will call the callback:

``` javascript
jsonld.flatten(doc, null, function(err, flattened) {
  // all deep-level trees flattened to the top-level
});
```
### Make `ctx` optional

``` javascript
jsonld.flatten = function(input, ctx, options, callback) {
  //...
  if(typeof options === 'function') {
    callback = options;
    options = {};
  }
  // add this part
  else if(typeof ctx === 'function') {
    callback = ctx;
    ctx = null;
    options = {};
  }
```

I prefer the latter, or some perturbation (i.e. `argument[arguments.length - 1]`) since the callback is _not_ optional, correct? `null` seems like a pretty reasonably, if boring, default, and could be assumed given only two inputs... and aligns with what is documented, which must have been intended at some point.

I will prepare a pull request, if desired: would it also make sense to add some tests for these API usage things like this? The EARL testing machinery is a little imposing :)

PR #43, opened 2013-10-21
API tests

bollwyvl:
After discussion on #42, decided to give this a whack.

This is a start, and should not be considered the _actual_ PR, as I'd like to find out the maintainers' thoughts on this approach. Here I am referring to the EARL stuff, and whether I need IDs and all the other stuff to be doing it "right." 

This set of tests doesn't do much to coverage yet at this point, as they really only hit a few lines of argument count checking... the error case of too few arguments will never be correctly executed, as `callback` is usually required, which by definition can't be populated if the argument count is 0.

Also, I am not exercising any of the `options`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,26 +4,30 @@
   "description": "A JSON-LD Processor and API implementation in JavaScript.",
   "homepage": "http://github.com/digitalbazaar/jsonld.js",
   "author": {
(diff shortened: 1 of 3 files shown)


dlongley:
This is likely too out-of-date to ever land and we've made a number of improvements to the test suite that probably cover it, closing.

PR #68, opened 2014-08-12
Patch fix for aggressive embed support in framing algorithm

### LINKED ISSUE #20: aggressive re-embedding
Also just tried Markus' version of playground and it works
http://www.markus-lanthaler.com/jsonld/playground/

....
 Just found that this is discussed in  https://github.com/json-ld/json-ld.org/issues/119
So it is addressed?

.....
If I've an object  that will be referenced by multiple places, seems that only the 1st reference will be embedded and the rest wont be...

Here is my examples that I tried out in the playground... 
ID_0 object has  two objects. ID_1 and ID_2.

Then both ID_1 and ID_2, will have one field shared, ID_SHARE_1... 
Then, if I've a frame as 

---> Frame
```
{
"@context": {"C" : "http://c" },

"@type": "http://root",
"C:has_items" : {}
}

```
--> framed output... Shouldn't ID_SHARE_1 object should be embedded in both ID_1 and ID_2 objects??? 

```
{
  "@context": {
    "C": "http://c"
  },
  "@graph": [
  {
    "@id": "http://ID_0",
    "@type": "http://root",
    "C:has_items": [
    {
      "@id": "http://ID_1",
      "C:has_shared_item": {
        "@id": "http://ID_SHARE_1",
        "C:hasName": "I'm shared"
      }
    },
    {
      "@id": "http://ID_2",
      "C:has_shared_item": {
        "@id": "http://ID_SHARE_1"
      }
    }]
  }]
}
```

---->> Raw Input

```
{
"@context": {"C" : "http://c" },

"@graph": [

  {
  "@id" : "http://ID_SHARE_1",
  "C:hasName": "I'm shared"
  },

  {
   "@id": "http://ID_1",
   "C:has_shared_item" : 
    {
      "@id": "http://ID_SHARE_1"
    }
  },
  {
   "@id": "http://ID_2",
   "C:has_shared_item" : 
    {
      "@id": "http://ID_SHARE_1"
    }
  },

  { 
  "@id": "http://ID_0",
  "@type": "http://root",
   "C:has_items": [
      {"@id": "http://ID_1"},
      {"@id": "http://ID_2"}
  
    ]
  }
]
}
```

julianeverett:
With reference to jsonld framing algorithm at http://json-ld.org/spec/latest/json-ld-framing/, this fixes #20 via removal of 4.4.4 and addition of checks for circular reference injection prior to the recursive calls at 6.2.3.1.1 and 6.2.3.

### DIFF
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -3986,7 +3986,6 @@ function _frame(state, subjects, frame, parent, property) {
 
   // add matches to output
   var ids = Object.keys(matches).sort();
-
   for(var idx in ids) {
     var id = ids[idx];
 
@@ -4016,15 +4015,12 @@ function _frame(state, subjects, frame, parent, property) {
     // iterate over subject properties
     var subject = matches[id];
     var props = Object.keys(subject).sort();
-
     for(var i = 0; i < props.length; i++) {
       var prop = props[i];
 
       // copy keywords to output
       if(_isKeyword(prop)) {
-
         output[prop] = _clone(subject[prop]);
-
         continue;
       }
 
@@ -4039,7 +4035,6 @@ function _frame(state, subjects, frame, parent, property) {
 
       // add objects
       var objects = subject[prop];
-
       for(var oi = 0; oi < objects.length; ++oi) {
         var o = objects[oi];
 
@@ -4076,7 +4071,6 @@ function _frame(state, subjects, frame, parent, property) {
 
     // handle defaults
     var props = Object.keys(frame).sort();
-
     for(var i = 0; i < props.length; ++i) {
       var prop = props[i];
 


kevinSuttle:
Is this behavior documented and/or available at json-ld.org/playground/?

dlongley:
@kevinSuttle, it's been live on the playground for a while now. It's poorly documented, but live. I imagine `@link` doesn't work very well given that it creates cycles, but `@last`, `@always`, and `@never` can be set as values for `@embed`.

...I just checked and it seems like `@last` might be buggy, though.

kevinSuttle:
@dlongley Cool. Are there docs for how this should work in practice?

PR #79, opened 2014-11-23
simplify instructions on using jsonld cli tool

ralphtheninja:


davidlehn:
Tool moved to https://github.com/digitalbazaar/jsonld-cli and changes applied to that README.

PR #94, opened 2015-02-24
Correct documentation of documentLoader usage.

alexdeleon:
A small correction to the README.

At least on the browser (not sure if it is the same behaviour in node.js) the documentLoader returns a promise. Passing the callback does nothing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,13 @@ var customLoader = function(url, callback) {
         documentUrl: url // this is the actual context URL after redirects
       });
   }
-  nodeDocumentLoader(url, callback);
+  nodeDocumentLoader(url).
+    then(function(resp){
+      callback(null, resp);
+    }).
+    catch(function(err){
+      callback(err, null);
+    });
 };
 jsonld.documentLoader = customLoader;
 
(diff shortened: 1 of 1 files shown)


dlongley:
The node.js document loader uses the callback. The promise-API is preferred in the browser.

alexdeleon:
I documented the difference between node.js and browser interfaces. What do you think?

dlongley:
That's better, but, to be a bit more accurate:

By default the node document loader won't use promises (callbacks are the preferred mechanism in node) and, by default, the browser-based document loaders (xhr and jquery) will use promises *if* they are found in the global environment (they are native or polyfilled -- promises are preferred in the browser). So maybe just add some text that says "By default, if Promises are supported or polyfilled in the browser ...". That and/or a sentence that indicates you can use a callback-based API or a promise-based API via options with a quick parenthetical about the defaults would be good.

You can see the docs in the code here:

https://github.com/digitalbazaar/jsonld.js/blob/master/js/jsonld.js#L1644-L1645
https://github.com/digitalbazaar/jsonld.js/blob/master/js/jsonld.js#L1776-L1778
https://github.com/digitalbazaar/jsonld.js/blob/master/js/jsonld.js#L1556-L1558

alexdeleon:
Thanks @dlongley, that's a very clear explanation. I have updated the README file with your suggestions.

dlongley:
Looks, great, @alexdeleon. Thanks!

PR #107, opened 2015-08-04
Fix "setImmediate" bug for browserify and IE10

ariutta:
Issue: using jsonld.js in a browserify project and attempting to call the `normalize` method resulted in IE10 throwing the error "Invalid calling object".

It's the same as described in this issue for a different library (async.js): https://github.com/caolan/async/pull/350

Solution: handle `setImmediate` and `nextTick` the same way async.js does.


dlongley:
Landed as 0bcc971dd3a7c8d6300267db6b26689fa86e4eda.

PR #109, opened 2015-08-16
Also ignore pkginfo when doing a browser build

warpr:
This seems to be necessary when using ```require("jsonld")``` in a project which builds a browser bundle using webpack.


davidlehn:
Thanks.  Avoided merge conflict and added in https://github.com/digitalbazaar/jsonld.js/commit/e5a1dbc28ad618387d323110c98b7bb8a8718086.

PR #113, opened 2015-09-13
Ignore null values in language maps

nichtich:
The spec explicitly allows null values in language maps (section 8.5).


nichtich:
Related to json-ld/json-ld.org#375.

dlongley:
Thanks @nichtich!

PR #122, opened 2016-01-05
N-Quads may contain comments. Fixes #121

### LINKED ISSUE #121: N-Quads parser doesn't allow comments
The [N-Quads parser](https://github.com/digitalbazaar/jsonld.js/blob/master/js/jsonld.js#L6816-L6933) doesn't allow the parsed data to contain comments. As the [N-Quads specification](http://www.w3.org/TR/n-quads/#h2_sec-grammar) says: 

> Comments in N-Quads take the form of '#', outside an IRIREF or STRING_LITERAL_QUOTE, and continue to the end of line (EOL) or end of file if there is no end of line after the comment marker. Comments are treated as white space.

However, if I try to parse the following N-Quads (or any other N-Quads containing comments) using jsonld.js, it throws `jsonld.ParseError: Error while parsing N-Quads; invalid quad.`:

```
# NQuads
<http://example.com/resource/person/1> <http://xmlns.com/foaf/0.1/name> "Person 1" <http://example.com/resource/graph/1> .
```

Can the parser be extended to allow comments?

jindrichmynarz:
RDF serialized in N-Quads may contain comments. Comments starting with the `#` character are ignored.

### DIFF
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -6828,10 +6828,11 @@ function _parseNQuads(input) {
   var datatype = '(?:\\^\\^' + iri + ')';
   var language = '(?:@([a-z]+(?:-[a-z0-9]+)*))';
   var literal = '(?:' + plain + '(?:' + datatype + '|' + language + ')?)';
+  var comment = '(?:#[^$]*)?';
   var ws = '[ \\t]+';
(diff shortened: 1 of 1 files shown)


dlongley:
Thanks @jindrichmynarz!

PR #126, opened 2016-01-27
Modularize code

ariutta:
The URL parser had a `full` and a `simple` option, but the `simple` option was never used. Removing the unused `simple` option will speed up the code and make the total file size smaller.

I also pulled `jsonld.url.parse` and `_removeDotSegments` into their own modules to make it easier to read the main jsonld.js file. ~~These modules are currently referenced in `package.json` under my github account, but I'd be happy to transfer these repos to digitalbazaar or the jsonld.js organization.~~


ariutta:
Note that removing `jsonld.url.parsers` and removing the `parser` argument from `jsonld.url.parse` means a slight change to the public API.

dlongley:
In the near future, we're likely going to break jsonld.js into a number of different files and then add a build system to cobble them back together. Our plan is to do this all at once, but it's unclear what will get its own module and what will just be a file here in this repo. There are bits that definitely should get their own module (eg: we're working on this https://github.com/digitalbazaar/rdf-canonize/tree/dev), but it's doubtful we'll break out things like the `_removeDotSegments` into separate modules because of the maintenance overhead that generates.

ariutta:
@dlongley: sure, I understand the maintenance overhead. The Babel project uses a monorepo that contains multiple NPM packages. Here's a description of how they do it and the tool they created to manage this: https://github.com/kittens/lerna

I undid the splitting out of `_removeDotSegments`. Are you onboard with the other changes?

ariutta:
I really need to minimize the file size for browser usage. That's why I removed the unused `simple` option from the URL parser.

dlongley:
The `nextTick` changes won't work because it should be using `setImmediate` when it's available to allow I/O to be processed during normalization of large/complex datasets. However, we're separating out the normalization/canonization stuff entirely right now so we can probably just leave that alone and wait for those changes to be done.

Regarding the url parser change, including another library using a synchronous `require` will break existing AMD setups.

We really need to get a build system in place to cobble together the bits people need so people can pick and choose what they want if they'd like to create custom builds to cut down on file size for the browser. Again, people may not need normalization/canonization at all (or maybe they don't need framing) -- and those will result in far better savings in terms of size than a few lines in the URL parser.

ariutta:
Replaced `nextTick` polyfill with `setImmediate` polyfill.

Also added a minimal build system that creates `./dist/jsonld.js` via a `postinstall` lifecycle script. The build can also be run from the command line with `npm run build`.

ariutta:
The build system generates a bundle that has the `umd` format:

> `umd` – Universal Module Definition, works as `amd`, `cjs` and `iife` all in one
[from the Rollup docs](https://github.com/rollup/rollup/wiki/JavaScript-API#format)

ariutta:
It's passing the node tests but not the browser tests at the moment.

ariutta:
You'll notice I made a proposed draft version of a modularized system with control over building.

ariutta:
Even though the browser tests are failing, it actually works in the browser. You can take a look at your browser console when you visit this test page and see results from test.js:

http://plnkr.co/edit/jkiQVUQBbFdZi2UoxGJQ?p=preview

ariutta:
## Test Statuses

**node tests**: all pass

**browser tests**: appear to be failing only due to a bug with the test code, not because of an actual bug in the library. It appears to work fine in the browser. You can see it running every example from the main README if you visit this test page and take a look at your browser console to view the results from test.js:

http://plnkr.co/edit/jkiQVUQBbFdZi2UoxGJQ?p=preview

**node-document-loader tests**: fail, because this version is using [superagent](https://github.com/visionmedia/superagent) and [superagent-cache](https://github.com/jpodwys/superagent-cache) instead of the existing node document loader. This change allows for re-enabling caching in both node and the browser.

ariutta:
The more recent pushes implement a proposed version of a modularized system with control for customized builds. This system supports [tree shaking](https://medium.com/@Rich_Harris/tree-shaking-versus-dead-code-elimination-d3765df85c80#.wnmup84n3), which "only includes the bits of code your bundle actually needs to run," as implemented by [Rollup](https://github.com/rollup/rollup). If a user only needs to expand data, tree shaking allows that user to pull in just the code needed to run `jsonld.expand`, skipping all the extra unused code for compacting, framing, normalization, etc.

The build system by default creates `./dist/node/jsonld.js` and  `./dist/node/browser.js` via a `postinstall` lifecycle script. These files can also be generated from the command line with `npm run build`. The `rollup.config.*.js` files are where the build preferences are set. It would be straightforward to add a build option to allow a user to specify only the parts of the jsonld API that they need in a custom build.

For the browser, the build system generates bundles that have the `umd` format:

> `umd` – Universal Module Definition, works as `amd`, `cjs` and `iife` all in one
[from the Rollup docs](https://github.com/rollup/rollup/wiki/JavaScript-API#format)

davidlehn:
We did want to be more modular, but **~128** files is a level far beyond modular!  Individual files for single constants is an unmaintainable coding style, don't you think?  There are tradeoffs between importing just what you need and making it crazy difficult to edit the code.  I think the approach here is on the difficult to use side.  How many bytes is this saving in a common use case where someone is using many of the top level API functions?

I assume the reason for this file split up is to leverage tree shaking of ES2015 tools.  That is a good goal, but rollup and similar tools do do some dead code elimination too.  Have you tried using larger files with many exports and seeing how well they get optimized when run through, say, rollup with uglify-js?

The normalization code is already in the process of breaking out into a different module entirely (for various reasons):
https://github.com/digitalbazaar/rdf-canonize/tree/dev

Another issue is that I don't think we want to use ES2015 modules in jsonld.js at this time.  I think our perspective at Digital Bazaar is to keep the code able to run with default LTS nodejs options for ease of development and wider compatibility.  For a number of our projects, including jsonld.js, we are exploring various tools (rollup, webpack, browserify, etc) to build browser compatible blobs out of a CommonJS based module style.  As nodejs and v8 gain more mature ES2015 module support and usage guidelines we'll consider switching over.

ariutta:
Regarding the modules for individual variables, definitely agree that's
overkill. This is a first draft for starting discussion,  and as a first
cut, I just did an export for every top-level variable and function
declaration. Now we can decide where to more appropriately draw the line.

Regarding ES2015, this version is actually ES5 for Node and the browser.
Look at the "main" property in package.json. The ES2015 code is for the
"esnext:main" property in package.json.

ariutta:
Moved all top-level constants into a single file.

ariutta:
Size comparisons:

* existing version, minified: 120758
* modularized, minified full version as in plunkr demo: 90487
* modularized version with just `jsonld.expand`: 48153

ariutta:
Added custom bundling feature so users can choose which modules to include:
```
rollup -c rollup.config.custom.js --modules=compact,expand
```
=> ./dist/custom/jsonld.js

ariutta:
Before continuing with this, I'd love to make sure everyone's on board with the direction it's going. To me, it seems clear that the bundle size reduction and custom bundle creation capability make the library much more amenable to in-browser usage, but I realize this is a big change to the organization of the codebase.

I also have a couple of specific questions. There are still some improvements to be made, especially for cases like how the `Processor`/`JsonLdProcessor` is modularized:

* https://github.com/ariutta/jsonld.js/blob/master/lib/Processor.js
* https://github.com/ariutta/jsonld.js/blob/master/lib/JsonLdProcessor.js
* https://github.com/ariutta/jsonld.js/blob/master/lib/_declobberedJsonLdProcessor.js

Should the contents of those files be merged into one module? Or should we split up the `Processor` to help with minimizing bundle size, e.g., so that a user who only needs `jsonld.expand` does not also get the `Processor` code for compaction? Maybe the `Processor` code for each method should be merged with the related `jsonld` method, e.g., the code for `Processor.prototype.expand` could be in the `expand` module?

How should this version integrate with the [external normalization module](https://github.com/digitalbazaar/rdf-canonize/tree/dev)?

The default value for `maxRedirects` [is none](https://github.com/digitalbazaar/jsonld.js/blob/c43f112df04141be81fe468dc582ec90efe8d68f/js/jsonld.js#L1735). But [some of the tests](https://github.com/json-ld/json-ld.org/blob/b2c57989c1d72faf5d6fe788c0523120d8098136/test-suite/tests/remote-doc-manifest.jsonld#L47) expect that redirects are followed. Where is `maxRedirects` set for those tests?

dlongley:
@aruitta,

> Before continuing with this, I'd love to make sure everyone's on board with the direction it's going.

Thank you for all of this effort. Here's some feedback on the changes here with respect to the direction we'd like to see the library head in:

Good:

* Splitting up the monolithic jsonld.js file
* Ability to only pull in the basic features that you need
* Build system that can generate UMD

Needs work:

* We do not want the module API to be ES2015 yet. It should be in CommonJS format, at a minimum. The reason for this is so that we can load it in node without any build steps at all. We need at least one dev environment (node being the most commonly used one for dev) to not require any build steps. The introduction of a build step is too much of a hindrance to rapid development. It's fine if there's a build step to generate code to run in the browser/custom browser bundles.
* Code is now 100+ files, which is far too many to reasonably maintain or understand. To reduce the number of files, the code should be organized according to features and concepts, not individual functions, constants, and so forth.  For example, a single file for a class like `IdentifierIssuer` or `Permutator` is good (but more on these specific examples below), a single file for each JSON-LD API operation and associated helpers is good (so like, one for expansion, one for compaction, one for framing, etc). This is how people will end up recombining the library anyway -- by picking according to the features they need.
* We should pull in the external canonicalization module (rdf-canonize). We should not duplicate efforts (we're splitting up canonicalization according to concept/feature in that repo already and we want to reuse that code). However, that module is incomplete, so whatever CommonJS-based reorganization and build system we figure out over here, we can copy over to that module and then finish it up. That module will be completed before this PR is accepted (it is also a higher priority to complete, other groups are waiting on that work). Then we can figure out how to integrate it here.
* Regarding the processor, a core file should exist that defines the processor without any functionality that file can then be required by separate files that attach each JSON-LD API feature (expansion, compaction, framing).
* Take a look at how the code is broken up conceptually in [rdf-canonize](https://github.com/digitalbazaar/rdf-canonize/tree/dev/lib), there's really no reason to use any more files for breaking up the normalization/canonicalization feature than that.

> The default value for maxRedirects is none. But some of the tests expect that redirects are followed. Where is maxRedirects set for those tests?

That will have to be investigated ... maybe it isn't at the moment and that will need to be addressed.

dlongley:
@ariutta,

> Maybe the Processor code for each method should be merged with the related jsonld method, e.g., the code for Processor.prototype.expand could be in the expand module?

To be clear, the above is what I was suggesting we do. There should be a single file for the "expansion feature" that includes all of the above.

ariutta:
Good feedback. I'll continue working, but I may slow down until `rdf-canonize` is completed so I can pull that code in without duplicating effort.

One more question: what API do we want to expose for users to import/require specific modules like `jsonld.expand` in their own code? The style `lodash` uses is easy to consume, but the build step is more complicated:

```js
var _assign = require('lodash/assign');
```

The way it works is that when you run `npm install`, there's an automatic build step that creates, in the top-level `lodash` directory, a huge number of files like `./node_modules/lodash/assign.js`. But the actual source code appears to be all in [one big file](https://github.com/lodash/lodash/blob/62962791b3fbc7e68655095d9caf919efe1cebc8/lodash.js#L11187). I propose we use one of the following:

```js
var expand = require('jsonld/expand');
// vs.
var expand = require('jsonld/lib/expand');
```

The first might mean we need a build step like lodash uses. The second is uglier, but might be easier to set up.

leipert:
@ariutta @dlongley 
What is the state of this. How can I help?

dlongley:
@leipert -- thanks for offering to help. I think this work has stalled, so feel free to what you can. My comment here lays out what we're looking for:

https://github.com/digitalbazaar/jsonld.js/pull/126#issuecomment-193940445

dlongley:
@leipert,

I'll add a caveat that we don't necessarily need to complete rdf-canonize before the reorganization happens here... and we may find that we can finish both off based on the efforts here. rdf-canonize is essentially just waiting on a good module infrastructure/framework to copy from this library. It has the same problem, except it has been split up already -- it just needs the capability to be sewn back together.

ariutta:
My work was really to spark discussion and get something going, but it is
not production ready. I'm not going to be able to get back to this anytime
soon, so feel free to use anything useful and ignore everything else from
my refactoring.

BigBlueHat:
First, thanks to @ariutta for taking the first crack at this!

Second, I've been reading up on [System.js](https://github.com/systemjs/systemjs#systemjs) which @tilgovi's suggested would be a good fit for [Apache Annotator](http://annotator.apache.org/).

I'm not keen on chasing the "new shiny thing" but it seems these folks have been at this for awhile, it has a pretty proven track record and can do some fabulous async loading and/or bundling and the like.

Would there be any interest in a PR *attempt* at system.js usage for this repo?

Cheers!
🎩

dlongley:
@BigBlueHat, @davidlehn is working on this problem right now ... so don't do anything just yet without him chiming in. :)

dlongley:
Closing this is out-of-date -- we long ago did a rewrite to better modularize the code. Any further refactoring should be done via a new PR.

Issue #130, opened 2016-02-26
Authorization with node documentLoader


agibson-vrnt:
Can't find support anywhere for setting the Authorization header in the node.js documentLoader. Is this supported by the library / are you open to pull requests?

dlongley:
It's not supported but we're open to PRs. One way to do it would be to add a `headers` option here:

https://github.com/digitalbazaar/jsonld.js/blob/master/js/jsonld.js#L1728-L1742

And let people set whatever headers they want -- and then merge in the default `Accept` header on top of any custom headers in the code below.

goofballLogic:
I see suites of tests for the json-ld algorithms, but it's not clear how I add tests for this change which is more to do with the workings of this library than the algorithms. Where do I add the tdd/bdd for this change?

dlongley:
@goofballLogic,

The test suite for jsonld.js needs to be reworked so that it has better support for both spec compliance and local tests. It would be easiest to have two separate test suites (one for each of the aforementioned) because they work quite differently from each other. We just haven't been able to prioritize that work.

So your options are to:

1. Add the non-test parts but leave this issue open indicating we need test(s) for it.
2. Take a stab at setting up a separate `test` dir and mocha framework for running local tests and add your test as the first one.

goofballLogic:
Do you release to npm on a specific cadence? Just wondering when we should expect to get the code in the pull request covering this issue.

dlongley:
@goofballLogic,

Released in 0.4.6: ec73a78d9d51fb032f44cf010a0f0afb5b7c683b

goofballLogic:
awesome.




---- On Wed, 02 Mar 2016 20:33:54 +0000 Dave Longley &lt;notifications@github.com&gt;wrote ---- 




@goofballLogic,

Released in 0.4.6: ec73a78

PR #132, opened 2016-02-29
Added passing headers as part of the options hash for node document loader

goofballLogic:
0. node document loader can now accept "headers" as part of the options object.
    0. if `Accept` or `accept` are specified as part of headers, a `RangeError` is thrown specifying that only 
    `"Accept header may not be specified as an option; only "application/ld+json, application/json" is supported"`
    0. if `Accept` or `accept` are not specified, or the options object isn't specified, the default header of `application/ld_json, application/json` is passed
0. To make these changes more robust, I've added mocha tests in /test:
    0. To run local tests (ones that aren't testing the JSON-LD algorithms), use any of the following:
        - `npm run test-local`
        - `make test-local`
        - `mocha`
0. **[ N.B. side-effect ]**: to allow for efficient testing, I've added a facility to inject the "request" object as part of the options object, so that one can mock the interface of request.js. The corollary is that this facility could be used to inject one's own implementation of request.js at run-time. To avoid this we would need to spin up a local service rather than mocking.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,10 @@ test-coverage:
 	./node_modules/.bin/istanbul cover ./node_modules/.bin/_mocha -- \
 		-u exports -R $(REPORTER) $(TESTS)
 
+test-local:
+	./node_modules/.bin/mocha
+
 clean:
 	rm -rf coverage
 
-.PHONY: test test-node test-browser test-local-node test-local-browser test-normalization-node test-normalization-browser test-coverage clean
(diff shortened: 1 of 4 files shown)


dlongley:
Thanks! Landed as bc38da02d1953100806c66ca9950ead25327ab60.

goofballLogic:
Sweet. How do you schedule sending updates to npm ?

dlongley:
Addresses #130.

Issue #141, opened 2016-05-02
Wrong MIME type for N-Quads


miguel76:
According to the W3C Recommendation, the standard MIME type for N-Quads is "application/n-quads" (see https://www.w3.org/TR/n-quads/#h2_sec-mediaReg), while jsonld currently requires "application/nquads" in `.normalize()`, `.toRDF()` and `fromRDF()`.
I propose it should accept both "application/n-quads" and "application/nquads", returning N-Quads in both cases.

davidlehn:
Another attempt at this: https://github.com/digitalbazaar/jsonld.js/pull/223

davidlehn:
Fixed in 0.5.18:
https://github.com/digitalbazaar/jsonld.js/commit/eef1932e96889daa569199d3525bf7ec83119143

PR #148, opened 2016-08-03
Bug/support nquads

ivikash:
Hey, 

As pointed out in #141, This PR intends to support both `application/nquads` and `application/n-quads` in jsonld.js.

There are a few things, that I wanted to know
- Should we still support `application/nquads`?
- In this PR I am using `options.useStandardNQuadsType` and if the user passes `true` then only `application/n-quads` formatting is done. Should I make this behaviour by default? And also should I change the variable name to something more generic?

### DIFF
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -756,9 +756,9 @@ jsonld.objectify = function(input, ctx, options, callback) {
  *          [base] the base IRI to use.
  *          [expandContext] a context to expand with.
  *          [inputFormat] the format if input is not JSON-LD:
- *            'appilcation/n-quads' for N-Quads
+ *            'appilcation/n-quads' for N-Quads.
  *          [format] the format if output is a string:
- *            'appilcation/n-quads' for N-Quads
+ *            'appilcation/n-quads' for N-Quads.
  *          [documentLoader(url, callback(err, remoteDoc))] the document loader.
  * @param callback(err, normalized) called once the operation completes.
  */
@@ -787,8 +787,13 @@ jsonld.normalize = function(input, options, callback) {
     options.documentLoader = jsonld.loadDocument;
   }
 
-  if('inputFormat' in options) {
-    if(options.inputFormat !== 'application/n-quads') {
+  var opts = _clone(options);
+
+  if('inputFormat' in opts) {
+    if(opts.inputFormat.indexOf('application/nquads') != -1) {
+      opts.inputFormat = 'application/n-quads';
+    }
+    if(opts.inputFormat !== 'application/n-quads') {
       return callback(new JsonLdError(
         'Unknown normalization input format.',
         'jsonld.NormalizeError'));
@@ -798,7 +803,6 @@ jsonld.normalize = function(input, options, callback) {
     new Processor().normalize(parsedInput, options, callback);
   } else {
     // convert to RDF dataset then do normalization
-    var opts = _clone(options);
     delete opts.format;
     opts.produceGeneralizedRdf = false;
     jsonld.toRDF(input, opts, function(err, dataset) {
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -756,9 +756,9 @@ jsonld.objectify = function(input, ctx, options, callback) {
  *          [base] the base IRI to use.
  *          [expandContext] a context to expand with.
  *          [inputFormat] the format if input is not JSON-LD:
- *            'application/nquads' or 'application/n-quads' for N-Quads.
+ *            'appilcation/n-quads' or 'application/n-quads' for N-Quads.
  *          [format] the format if output is a string:
- *            'application/nquads' or 'application/n-quads' for N-Quads.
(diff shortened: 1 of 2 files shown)


js/jsonld.js
```diff
@@ -788,7 +788,7 @@ jsonld.normalize = function(input, options, callback) {
   }
 
   if('inputFormat' in options) {
-    if(options.inputFormat !== 'application/nquads') {
+    if((options.inputFormat !== 'application/nquads') || (options.inputFormat !== 'application/n-quads'))  {
```
dlongley: Shouldn't this be an `&&`? Otherwise it would seem that this compound conditional would always be true.

That being said, it would probably be cleaner to make a shallow copy of the options and if the `inputFormat` is `application/nquads` it should just be changed early to the correct `application/n-quads`. Then we only need to check for that in the code later.

js/jsonld.js
```diff
@@ -821,6 +821,8 @@ jsonld.normalize = function(input, options, callback) {
  * @param [options] the options to use:
  *          [format] the format if dataset param must first be parsed:
  *            'application/nquads' for N-Quads (default).
+ *          [useStandardNQuadsType] the format of N-Quads
+ *            (default: false)
```
dlongley: I don't think we need this option. All we're really doing is correcting for what could now be seen as a typo ... so if someone uses `application/nquads` for the `inputFormat`, let's just correct it for them and then proceed as if the proper format was given.

ivikash:
@dlongley So, I have made the necessary changes. Yes, indeed you were right we should just convert `nquads` to `n-quads` and infact that was something I was willing to raise from point 1. Also, now inside both `if..else` block we are making a shallow copy of `options` as `opts` as a result, I have moved it outside the block :)

ivikash:
And why these travis tests are failing?

js/jsonld.js
```diff
@@ -787,8 +787,13 @@ jsonld.normalize = function(input, options, callback) {
     options.documentLoader = jsonld.loadDocument;
   }
 
-  if('inputFormat' in options) {
-    if(options.inputFormat !== 'application/nquads') {
+  var opts = _clone(options);
+
+  if('inputFormat' in opts) {
+    if(opts.inputFormat.indexOf('application/nquads') != -1) {
```
dlongley: Comparison should do:

`opts.inputFormat === 'application/nquads'`

js/jsonld.js
```diff
@@ -918,7 +922,7 @@ jsonld.fromRDF = function(dataset, options, callback) {
  *          [base] the base IRI to use.
  *          [expandContext] a context to expand with.
  *          [format] the format to use to output a string:
- *            'application/nquads' for N-Quads.
+ *            'appilcation/n-quads' for N-Quads.
```
dlongley: Typo.

js/jsonld.js
```diff
@@ -959,7 +963,7 @@ jsonld.toRDF = function(input, options, callback) {
       // output RDF dataset
       dataset = Processor.prototype.toRDF(expanded, options);
       if(options.format) {
-        if(options.format === 'application/nquads') {
+        if(options.format === 'appilcation/n-quads') {
```
dlongley: Typo.

js/jsonld.js
```diff
@@ -4005,7 +4009,7 @@ Normalize.prototype.main = function(dataset, callback) {
 
   // handle invalid output format
   if(self.options.format) {
-    if(self.options.format !== 'application/nquads') {
+    if(self.options.format !== 'appilcation/n-quads') {
```
dlongley: Typo.

js/jsonld.js
```diff
@@ -4217,7 +4221,7 @@ Normalize.prototype.main = function(dataset, callback) {
           normalized.sort();
 
           // 8) Return the normalized dataset.
-          if(self.options.format === 'application/nquads') {
+          if(self.options.format === 'appilcation/n-quads') {
```
dlongley: Typo.

js/jsonld.js
```diff
@@ -7022,7 +7026,7 @@ function _parseNQuads(input) {
 }
 
 // register the N-Quads RDF parser
-jsonld.registerRDFParser('application/nquads', _parseNQuads);
```
dlongley: We might actually need to keep this for backwards compatibility. I'm not sure.

js/jsonld.js
ivikash: Oh God! What was I thinking 😡 👊. my bad

js/jsonld.js
ivikash: I guess yes we should have both to support backward compatibility

davidlehn:
Due to significant code changes, I redid this patch.  Perhaps not addressing the "change the format early" issue though.
https://github.com/digitalbazaar/jsonld.js/pull/223

davidlehn:
Fixed in 0.5.18:
https://github.com/digitalbazaar/jsonld.js/commit/eef1932e96889daa569199d3525bf7ec83119143

Issue #156, opened 2016-12-01
Test failing on CI when re-running last build


harlantwood:
I forked the repo and ran tests on travis and got failures under all node versions: https://travis-ci.org/CoMakery/jsonld.js/builds/180316756

I suppose this must be that dependencies have changed in the 4 months since the build succeeded on the exact same code...

sample failures:

```
  1) JSON-LD JSON-LD Test Suite Framing #t0023 No match if node has a property where frame has an empty array for that same property.:
      AssertionError: { '@context': { ex: 'http://example.org/' }, '@graph': [] } deepEqual { '@context': { ex: 'http://example.org/' },
  '@graph': [ { '@id': 'ex:Sub1', 'ex:p': null, 'ex:q': 'bar' } ] }
      + expected - actual
       {
         "@context": {
           "ex": "http://example.org/"
         }
      -  "@graph": []
      +  "@graph": [
      +    {
      +      "@id": "ex:Sub1"
      +      "ex:p": [null]
      +      "ex:q": "bar"
      +    }
      +  ]
       }
      
      at Object.compareExpectedJson [as compare] (tests/test.js:468:12)
      at callback (tests/test.js:334:20)
      at js/jsonld.js:591:11
      at cleanup (js/jsonld.js:231:5)
      at js/jsonld.js:164:7
      at js/jsonld.js:2043:5
      at finished (js/jsonld.js:6780:7)
      at retrieve (js/jsonld.js:6786:7)
      at _retrieveContextUrls (js/jsonld.js:6885:3)
      at Function.jsonld.processContext (js/jsonld.js:2033:3)
      at js/jsonld.js:149:12
      at js/jsonld.js:133:9
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickCallback (internal/process/next_tick.js:98:9)
  2) JSON-LD JSON-LD Test Suite Framing #t0027 Recursively, replace all key-value pairs in compacted results where the key is @preserve with the value from the key-pair. If the value from the key-pair is @null, replace the value with null.:
     TypeError: Cannot use 'in' operator to search for '@omitDefault' in undefined
      at _getFrameFlag (js/jsonld.js:5073:23)
      at _frame (js/jsonld.js:5003:27)
      at Processor.frame (js/jsonld.js:3116:3)
      at js/jsonld.js:570:36
      at js/jsonld.js:367:7
      at finished (js/jsonld.js:6780:7)
[...]

PR #158, opened 2016-12-06
fixed the regex for bnodes when parsing n-quads.

Jimflip:
Actually only a partial fix to support '_' in the bnode label as per specification:

http://www.w3.org/TR/n-quads/#BNodes

```
The characters _ and digits may appear anywhere in a blank node label.
```

This causes issues on rdflib also which is utilising jsonld.js for json-ld serialisation.

### DIFF
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -6911,7 +6911,7 @@ if(!Object.keys) {
 function _parseNQuads(input) {
   // define partial regexes
   var iri = '(?:<([^:]+:[^>]*)>)';
-  var bnode = '(_:(?:[A-Za-z0-9]+))';
+  var bnode = '(_:(?:[A-Za-z0-9_]+))';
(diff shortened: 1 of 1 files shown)


Jimflip:
So the test doesn't build consistently before this PR.
So master is effectively broken?

dlongley:
@Jimflip,

> So the test doesn't build consistently before this PR.
> So master is effectively broken?

Some new framing tests were added recently due to some changes to the framing algorithm in the specification. We need to disable those for now so the tests pass again -- and then update our framing algorithm (when we have time) so they all pass. That's probably what you're seeing.

harlantwood:
This is probably also what we're seeing in #156.

Jimflip:
Just noticed this is still open, and no idea of how things have progressed since.

Shall I close this PR or have a crack at resolving conflicts?

PR #165, opened 2017-01-18
Fix the Mixed Content mess in XHR Requests

BigBlueHat:
Using https://www.w3.org/TR/upgrade-insecure-requests/

Added to the two browser-based document loaders: jquery & xhr

This avoids the Mixed Content errors when requesting non-https URLs which are also available over https. Fixes the security situation without changing the names...more or less.

Hope it's helpful!
🎩

### DIFF
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -1687,7 +1687,8 @@ jsonld.documentLoaders.jquery = function($, options) {
       },
       // ensure Accept header is very specific for JSON-LD/JSON
       headers: {
-        'Accept': 'application/ld+json, application/json'
+        'Accept': 'application/ld+json, application/json',
+        'Upgrade-Insecure-Requests': '1'
       },
       dataType: 'json',
       crossDomain: true,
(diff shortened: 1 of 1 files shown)


dlongley:
Thanks @BigBlueHat!

davidlehn:
How widely supported is this?  What happens if a client sends the header but can't handle the server response?
https://www.w3.org/TR/upgrade-insecure-requests/#preference
http://caniuse.com/#feat=upgradeinsecurerequests

BigBlueHat:
@davidlehn from what (little) I know about this, it seems that servers will typically issue a 307 redirect to the secure resource (as noted in the section you reference).

The CSP (server-side) declaration having an effect on the User Agent is the thing that would lack support in browsers--as in that case the UA is doing the rewriting.

In the case of the `Upgrade-Insecure-Requests: 1` client-sent header, the server (afaict) makes the choice of handling the preference (likely via 307's or similar).

But again. I'm no expert here...sadly.

BigBlueHat:
@davidlehn revert this at your leisure... CORS is in the way (afaict):
https://github.com/w3c/webappsec-upgrade-insecure-requests/issues/7

PR #169, opened 2017-02-13
Use npm-run-scripts for commandline tooling

BigBlueHat:
Useful for Windows development (hence the new `cross-env` dependency).

Makefile is still there in case deploy/host scripts depend on it...though
it could in turn use the npm run scripts.

The commands match the Makefile commands, so:
```bash
$ make test-node
$ # is the same as
$ npm run test-node
```

Let me know if it needs tweaks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "chai": "^3.5.0",
     "commander": "^2.8.0",
     "cors": "^2.7.1",
+    "cross-env": "^3.1.4",
     "express": "^4.13.3",
     "istanbul": "^0.4.3",
     "jscs": "^3.0.0",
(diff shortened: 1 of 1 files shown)


BigBlueHat:
Also, I should note that I've tested this in the Windows Linux Sub-System (lxss; aka "ubuntu in windows"), MingW, and PowerShell.

BigBlueHat:
Found and fixed a path issue when running in bash (mingw) in Windows. Should be good to go (completely) in all those scenarios now.

adambro:
Just an idea: Makefile will be just a way to run NPM scripts, so it could be replaced by [fakefile](https://github.com/kvz/fakefile)

dlongley:
I would like to close this out, I think #194 will supersede everything here -- so we shouldn't put any more effort into this PR. @davidlehn can you confirm?

davidlehn:
The Makefile will be gone.  Will have to check new commands for portability.  I'm guessing things like `${REPORTER:-spec}` that are sh specific might have issues.  So cross-env or a minor rework might be needed.  It's just nice to be able to do `SPEC=nyan npm t` when the need arises.

davidlehn:
The code has changed quite a bit.  This patch needs to be updated if it's still an issue.

Issue #172, opened 2017-04-23
Test suite is failing due to changes in the json-ld/json-ld.org repo.


mattcollier:
Tests are failing: https://travis-ci.org/digitalbazaar/jsonld.js/jobs/224895314

Due to: https://github.com/json-ld/json-ld.org/commit/7374dff787690ce6f7781b1079b9023acf550234

dlongley:
We need to do one of two things:

1. Finally completely rework and modernize our ancient test suite.
2. Hack something into the current ancient test suite to skip 1.1 tests.

cc: @davidlehn

davidlehn:
A first pass at skip logic was added a while ago:
https://github.com/digitalbazaar/jsonld.js/commit/41d432e379de9b49da5a516502008d9ac3426734
https://github.com/digitalbazaar/jsonld.js/commit/b79b80c87b1199430fbe0bf6b2d664f9ff68ebd1

It's currently a bad hack that skips all frame tests and url tests that were not passing due to the specs and tests being updated beyond what jsonld.js supports.  In theory all tests that are for 1.1 should be marked as such and the current code will skip them.

davidlehn:
Also see comment: https://github.com/json-ld/json-ld.org/commit/7374dff787690ce6f7781b1079b9023acf550234#commitcomment-21885156

dlongley:
Closing as addressed.

PR #176, opened 2017-04-27
Add headers option support to other loaders

lutangar:
Hi there,

Pretty self-explanatory, I didn't expect to find the `headers` options to be `node` loader only...
The support of this option is now extended to `jquery` and `xhr` loaders.

I may work on a `fetch` loader in the future, with the approriate polyfill for node, that would mean same code for node and the browser and a promise first API.

### DIFF
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -1646,6 +1646,41 @@ jsonld.cache = {
 };
 
 /**
+ * Accept header.
+ */
+var _defaults = {
+  headers: {
+    accept: 'application/ld+json, application/json'
+  }
+};
+
+/**
+ * Build an headers object from custom headers and assert `accept` header isn't overridden.
+ *
+ * @param {Object} optionsHeaders an object (map) of headers
+ *          with key as header name and value as header value.
+ * @return {Object} an object (map) of headers with a valid `accept` header.
+ */
+function buildHeaders(optionsHeaders) {
+  optionsHeaders = optionsHeaders || {};
+
+  var hasAccept = Object.keys(optionsHeaders).map(function(h) {
+    return h.toLowerCase();
+  }).indexOf('accept') !== -1;
+
+  if(hasAccept) {
+    throw new RangeError(
+      'Accept header may not be specified as an option; only "' +
+      _defaults.headers.accept + '" is supported.');
+  }
+
+  var headers = {'Accept': _defaults.headers.accept};
+  for(var k in optionsHeaders) { headers[k] = optionsHeaders[k]; }
+
+  return headers;
+}
+
+/**
  * Document loaders.
  */
 jsonld.documentLoaders = {};
@@ -1656,6 +1691,8 @@ jsonld.documentLoaders = {};
  * @param $ the jquery instance to use.
  * @param options the options to use:
  *          secure: require all URLs to use HTTPS.
+ *          headers: an object (map) of headers which will be passed as request
+ *            headers for the requested document. Accept is not allowed.
  *          usePromise: true to use a promises API, false for a
  *            callback-continuation-style API; defaults to true if Promise
  *            is globally defined, false if not.
@@ -1665,6 +1702,7 @@ jsonld.documentLoaders = {};
 jsonld.documentLoaders.jquery = function($, options) {
   options = options || {};
   var queue = new jsonld.RequestQueue();
+  var headers = buildHeaders(options.headers);
 
   // use option or, by default, use Promise when its defined
   var usePromise = ('usePromise' in options ?
@@ -1694,12 +1732,9 @@ jsonld.documentLoaders.jquery = function($, options) {
     $.ajax({
       url: url,
       accepts: {
-        json: 'application/ld+json, application/json'
+        json: _defaults.headers.accept
       },
-      // ensure Accept header is very specific for JSON-LD/JSON
-      headers: {
-        'Accept': 'application/ld+json, application/json'
-      },
+      headers: headers,
       dataType: 'json',
       crossDomain: true,
       success: function(data, textStatus, jqXHR) {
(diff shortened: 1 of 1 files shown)


js/jsonld.js
```diff
@@ -1646,6 +1646,30 @@ jsonld.cache = {
...
 
 /**
+ * Accept header.
+ */
+jsonld.acceptHeader = 'application/ld+json, application/json';
+
+/**
+ * Build an headers object from custom headers and assert Accept header is not override.
+ *
+ * @param headers an o
```
davidlehn: fix

js/jsonld.js
```diff
@@ -1646,6 +1646,30 @@ jsonld.cache = {
 };
 
 /**
+ * Accept header.
+ */
+jsonld.acceptHeader = 'application/ld+json, application/json';
```
dlongley: We should really have a `jsonld.defaults` object to put things like this under, if we don't have that already. If we don't have it, let's add it now before adding yet another new top-level thing:

```js
jsonld.defaults = {
  headers: {
    accept: 'application/ld+json, application/json'
  }
};
```

I wonder if it needs further scoping than that even. Thoughts from anyone else?

dlongley: If this isn't a default that could be changed or if we're not into figuring this out right now, let's not expose it in the public API.

lutangar: This isn't a default, I just reproduced the behaviour previously implemented on the node loader. 

But it in my opinion, it should only be default. The current implementation is far too restrictive on this `accept` header. The `RangeError` exception should only be raised when the `accept` header doesn't contain `application/ld+json` nor `application/json`.

This way it could be overridden for more exotic things like `application/ld+json, text/html, application/json`, that would still be valid.

What do you think?

dlongley: I recommend we just take `acceptHeader` off of the public API for now (off of  `jsonld.`) and make a private module defaults var:

```js
var _defaults = {
  headers: {
    accept: 'application/ld+json, application/json'
  }
};
```

That way we move towards something we can expose in the future but we don't have to decide on exactly what it's shape is since that's not the point of this PR right now.

js/jsonld.js
```diff
@@ -1646,6 +1646,30 @@ jsonld.cache = {
...
+
+/**
+ * Build an headers object from custom headers and assert `accept` header isn't overridden.
+ *
+ * @param {Object} headers an object of headers with key as header name and value as header value.
+ * @return {Object} an object of headers with a valid `accept` header.
+ */
+jsonld.buildHeaders = function (headers) {
+  headers = headers || {};
+  if ('Accept' in headers || 'accept' in headers) {
```
dlongley: Nits: Remove space after `if` and `function` to keep in the existing code style of this module.

Also, this check here will cover only two possible capitalizations of `accept`. It's annoying, but perhaps we should do this instead:

```js
var hasAccept = Object.keys(headers).map(function(h) {
  return h.toLowerCase();
}).indexOf('accept') !== -1;
if(hasAccept) {
  // throw range error
}
```

js/jsonld.js
```diff
@@ -1656,6 +1680,8 @@ jsonld.documentLoaders = {};
  * @param $ the jquery instance to use.
  * @param options the options to use:
  *          secure: require all URLs to use HTTPS.
+ *          headers: an array of headers which will be passed as request
```
dlongley: `headers` is actually not an array, but an object (map), right?

js/jsonld.js
dlongley: Yup, it's just a private marker for a half-baked idea.

lutangar:
Should I be concerned about these tests failing? It seems `master` also have these errors...
https://travis-ci.org/digitalbazaar/jsonld.js/jobs/226858807
```
1) JSON-LD JSON-LD Test Suite Error handling #tp002 If not specified using processingMode, processing mode is taken as json-ld-1.0:
    AssertionError: 'invalid term definition' == 'processing mode conflict'
2) JSON-LD JSON-LD Test Suite Error handling #tp003 If @version is specified, it must be 1.1:
    AssertionError: 'invalid term definition' == 'invalid @version value'
3) JSON-LD JSON-LD Test Suite Error handling #tp005 If not specified using processingMode, processing mode is taken as json-ld-1.0:
    AssertionError: 'invalid term definition' == 'processing mode conflict'
4) JSON-LD JSON-LD Test Suite Error handling #tp006 If @version is specified, it must be 1.1:
    AssertionError: 'invalid term definition' == 'invalid @version value'
```

dlongley:
@lutangar,

No those tests are an ongoing issue right not (#172), you can safely ignore them. It looks like the other tests pass, so you're good.

dlongley:
@davidlehn, can you please review to see if we're good to accept this? Thanks!

davidlehn:
`jsonld.buildHeaders` should probably just be a local function for now instead of being exposed on `jsonld`.  Does it need to be a public API?

`jsonld.buildHeaders` is mutating the input options and adding the 'Accept' field.  This would be a problem if you call the APIs with the same options object twice and the code checking for 'Accept' would throw an error the second time.  Can fix this by copying all the options headers over into a new headers object (like some of the code did before).

lutangar:
@davidlehn `buildHeaders` doesn't *need* to be public at the moment, this has been changed.
I also changed back headers construction to something closer to previous code, meaning without the mutation.

lutangar:
@dlongley @davidlehn is there anything else I could do help this to be merged?

dlongley:
@lutangar, I don't think so, it LGTM. @davidlehn, can we pull this in now?

davidlehn:
Looks ok.  I'll rebase and merge.  I'm going to add a patch to use some() vs map().  Not a big deal in this case but might as well: https://jsperf.com/check-for-lowercase-key/

PR #179, opened 2017-05-06
Self-provided test suite, skip two tests, and assert @version correctly when specified

aredridel:


davidlehn:
The test skipping part looks ok.

The `@version` check is maybe ok.  @dlongley: do we want to add in small parts of 1.1 support like this to 0.4.x?  Or should that just be in 0.5.x or is there some other plan?  Do we want to keep 0.4.x as just a json-ld-1.0 lib?  I do think that the test suite should be marked such that 1.0 libs would just skip tests that even needed this sort of patch.  Though we'll certainly need it for 1.1 support.

I'm not sure we want to merge in that test suite package change at the moment.  I kind of like the idea but I can see keeping that up-to-date will be an issue.  It's nice to just point at the latest test suite checkout and get new tests.  It's also a problem when the 1.1 tests get mixed in but we'll slowly deal with that.  There's also the issue that we, in theory, try to keep js, php, and python libs synced and they are all do a similar thing looking at the json-ld.org repo.  It's kind of nice to test them all with the same suite at the same time and it might be odd to use npm packages for python and php testing. ;-)  I'd say wait on that for a bit until we've considered it more.

davidlehn:
Thinking a bit more... the current way 1.1 tests are marked and skipped is problematic.  Opened a test suite issue on this.  Perhaps we can better skip all 1.1 related tests w/o listing specific tests.

davidlehn:
The 1.1 skipping is improved now since the test suite marks 1.1 tests with a specVersion property.

davidlehn:
`@version` support added in 0.5.x here:
https://github.com/digitalbazaar/jsonld.js/commit/f6a91c0642dbdfa29bf680085a8f5e6f5693e4b3

PR #180, opened 2017-05-06
Self-provided test suite and assert @version correctly when specified (port of #179 to 0.5.x)

aredridel:
This does not get CI working, as the tests (obviously) do not pass yet for this branch. However, it does make it possible to see real status in CI.
.

- I published the json-ld.org test suite as an npm module. I'm happy to pass on the keys to that npm package for updates if that is a concern; Also happy to maintain that going forward, I don't anticipate a lot of work there, but won't be making it a priority for my work unless someone wants to poke me about it
- I added code to conform to tp003 and tp006 tests, asserting that version is `1.1` if it is specified at all.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 TESTS = tests/test.js
 LOCAL_TESTS = test/*.js
 REPORTER = spec
+JSONLD_TEST_SUITE=node_modules/json-ld-test-suite
 
 all:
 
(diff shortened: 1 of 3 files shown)


davidlehn:
`@version` support added in 0.5.x here:
https://github.com/digitalbazaar/jsonld.js/commit/f6a91c0642dbdfa29bf680085a8f5e6f5693e4b3

gkellogg:
I think this PR is probably obsolete with the collective changes in https://github.com/digitalbazaar/jsonld.js/tree/0.6.x.

PR #182, opened 2017-05-12
Allow valid accept headers in document loader options

lutangar:
Hi guys,

Here is a feature that was previously mention in #176 
> The current implementation is far too restrictive on this accept header. The RangeError exception should only be raised when the accept header doesn't contain application/ld+json nor application/json.

>This way it could be overridden for more exotic things like application/ld+json, text/html, application/json, that would still be valid.

The point here is to allow valid headers to be used, because at the moment, even if the user specify exactly the default value (application/ld+json,  application/json) the `RangeError` is raised nonetheless.

Also, even if the `buildHeaders` function wasn't meant to be exposed, at the moment this is only way to unit test portions of code as there are no builds tools to allow to split things up and require them only where needed. By the way, is there any news/plans on this?

Anyway, what do you think of this feature?

### DIFF
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -1650,35 +1650,55 @@ jsonld.cache = {
  */
 var _defaults = {
   headers: {
-    accept: 'application/ld+json, application/json'
+    accept: ['application/ld+json', 'application/json']
   }
 };
 
 /**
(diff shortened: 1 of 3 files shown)


lutangar:
@dlongley @davidlehn any feedbacks on this?

Issue #167, opened 2017-02-07
Compacting with mapped @id and @type fails


jmatsushita:
From `0.4.10` compacted output with mapped `@id` and `@type` returns an `untitled` key.

Submitting failing test.

jmatsushita:
The updated test in json-ld/json-ld.org#464 fails with the following error message

```
  1) JSON-LD JSON-LD Test Suite Compaction #t0073 Ensure that compaction works with mapped @id and @type:

      AssertionError: { '@context': { id: { '@type': '@id', '@id': '@id' } },
  undefined: '../../.org/anno9',
  'http://example.org/property': 'blah' deepEqual { '@context': { id: { '@type': '@id', '@id': '@id' } },
  id: 'http://example.org/anno9',
  'http://example.org/property': 'blah
      + expected - actual

             "@type": "@id"
           }
         }
         "http://example.org/property": "blah"
      -  "undefined": "../../.org/anno9"
      +  "id": "http://example.org/anno9"
       }

      at Object.compareExpectedJson [as compare] (tests/test.js:515:12)
      at callback (tests/test.js:381:20)
      at cleanup (js/jsonld.js:231:5)
      at js/jsonld.js:164:7
      at js/jsonld.js:2045:5
      at finished (js/jsonld.js:6782:7)
      at retrieve (js/jsonld.js:6788:7)
      at _retrieveContextUrls (js/jsonld.js:6887:3)
      at Function.jsonld.processContext (js/jsonld.js:2035:3)
      at js/jsonld.js:149:12
      at js/jsonld.js:367:7
      at finished (js/jsonld.js:6782:7)
      at retrieve (js/jsonld.js:6788:7)
      at _retrieveContextUrls (js/jsonld.js:6887:3)
      at expand (js/jsonld.js:324:5)
      at done (js/jsonld.js:294:9)
      at Object.localLoader [as documentLoader] (tests/test.js:673:14)
      at js/jsonld.js:296:29
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickCallback (internal/process/next_tick.js:98:9)

```

davidlehn:
@jmatsushita What was the use case for this?  There's some discussion about how to fix it in https://github.com/digitalbazaar/jsonld.js/pull/183.

davidlehn:
Fixed via https://github.com/digitalbazaar/jsonld.js/pull/218.

PR #183, opened 2017-05-16
Fix compact-0073 test case.

davidlehn:
Related commit causing regression:
https://github.com/digitalbazaar/jsonld.js/commit/07e9c06dcae3ebb924a0eba6b5fcd9a76b702da2

Test case (added after the above commit):
https://github.com/json-ld/json-ld.org/commit/9268a3dfad927c70ccb6f1917697d0ac9885d7a5

This is a guess that fixes the test case.  Please review this patch as I'm not familiar with how this code and the inverseCtx works and the fix may need to be elsewhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 - Update url parser to remove default ports from URLs.
 - Skip spec version 1.1 tests.
 
+### Fixed
+- Fix compact-0073 test case.
+
 ## 0.4.12 - 2017-04-24
 
 ### Fixed
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -5498,7 +5498,11 @@ function _compactIri(activeCtx, iri, value, relativeTo, reverse) {
   // if term is a keyword, it can only be compacted to a simple alias
   if(_isKeyword(iri)) {
     if(iri in inverseCtx) {
-      return inverseCtx[iri]['@none']['@type']['@none'];
+      var type = inverseCtx[iri]['@none']['@type'];
+      if('@id' in type) {
+        return type['@id'];
+      }
+      return type['@none'];
     }
     return iri;
   }


dlongley:
This will only fix the narrow test case where `@type` has been defined to `@id` for a keyword alias. What if some other `@type` is specified like `foo:bar`?

I think the real problem is that `compact-0073` should probably produce an `invalid keyword alias` error. I don't know what the use case is for redefining the `@type` for a keyword -- even if it's technically "the correct type" as it is in the test case for `@id` and `@type` (these have an `@type` of `@id` intrinsically).

So I think perhaps the real fix here is to throw an `invalid keyword alias` error in `_createTermDefinition()` whenever any JSON key other than `@id` is present in an expanded keyword alias definition. So, if the value of `@id` in an expanded term definition is a keyword (indicating that a keyword alias is being created), *no other keys* can be present in that term definition or the `invalid keyword alias` error will be raised.

Thoughts? @davidlehn, If you agree, could you create an issue about this over at json-ld.org?

dlongley:
The alternative to throwing an error is to discard all of the additional properties in the expanded term definition when the `@id` property is a keyword in the appropriate place in `_createTermDefinition`.

I think the best place to do this would be after confirming that `@id` is in the term definition and we've expanded its value:

https://github.com/digitalbazaar/jsonld.js/blob/master/js/jsonld.js#L5864

If the `@id` value is a keyword, we should replace value with `{}` from that point forward. That would be a better patch for right now and one that is backwards compatible. If there's consensus via a json-ld.org issue that that processors should throw an error for this then we can change the line that blanks out `value` to raise that error.

davidlehn:
@gkellogg Any thoughts about this test case?

davidlehn:
Also see https://github.com/digitalbazaar/jsonld.js/issues/167.

gkellogg:
Note that the test has since been updated with the following context:

```
{
  "@context": {
    "id":      {"@type": "@id", "@id": "@id"},
    "type":    {"@type": "@id", "@id": "@type"}
  }
}
```

So, mapping is not an issue, and it seems clear that the expected result is reasonable. The question comes down to if it is an error for a term which maps to a keyword (such as `@id` or `@type` can have extra elements in the term definition, or should even be able to use a term definition. I don't see anything in the context processing which would through an "invalid keyword alias" if the term definition has keys other than `@id`, but this might be a reasonable thing to do in 1.1. However, there have been requests to be able to do something like `"@container": "@set"` for `@type` to get a more predictable shape.

For Expansion, the algorithm does not consider any `@type` mapping for `@type` or `@id`, so allowing this in the context might be unusual. If anything, it should complain if set to something else if not simply reject the presence of the `@type` key.

Dave's wording is mostly reasonable:

> Concretely: If the value of @id in an expanded term definition is a keyword (indicating that a keyword alias is being created), no other keys can be present in that term definition or the invalid keyword alias error will be raised.

However, this relates to https://github.com/json-ld/json-ld.org/issues/512, which perhaps need more comment. I can certainly see why you would want to be sure that the values of `@type` were predictably an array.

davidlehn:
Fixed via https://github.com/digitalbazaar/jsonld.js/pull/218.

Issue #168, opened 2017-02-07
Remove dependency on `pkginfo`


blakeembrey:
It seems pretty unneeded since you could just require the `package.json` yourself and expose the version. However, in my case, I'm having trouble using this module in the context of Runkit (see https://runkit.com/blakeembrey/json-ld-issue). The reason I'm trying to expose it there is that I'd really love to get a demo of https://github.com/blakeembrey/node-scrappy working and I'm using `jsonld` to expand the JSON before extracting the snippets.

Related to https://github.com/blakeembrey/node-scrappy/issues/13.

tolmasky:
Just a heads up, this because pkginfo is a peer dependency. If you require pkginfo you can get it working in RunKit today.

blakeembrey:
I'll give it a go. It's listed in dependencies for me though (https://github.com/digitalbazaar/jsonld.js/blob/0a7b8ba26b2d51c6cc5d784ce5e2d92ec13e4645/package.json#L29), unless we're looking at different things.

tolmasky:
Sorry for the delay, but this should work now. Was our thing.

davidlehn:
Fixed in https://github.com/digitalbazaar/jsonld.js/commit/17ec94a97a11576d319db643829f720db821d8a1 and released in 0.5.17.

Issue #152, opened 2016-09-02
Think about production dependencies


leipert:
Hey there.

I was just reading the code and wondered about a few dependencies/possible improvements:
- `pkginfo`: could be removed if the source code of [L8133-L8135](https://github.com/digitalbazaar/jsonld.js/blob/master/js/jsonld.js#L8133-L8135) is changed to:
  
  ```
  jsonld.version = require('../package.json').version
  ```

And this is somewhat related to #99:
- `request` is used for the node document loader. would it make sense to split json.ld into two modules, so that the json ld processor becomes a separate module, which is somewhat smaller and does not include the document loaders
- `xmldom` same as above, only used for rdfa converting, maybe provide it as a plugin

If you are somewhat happy about talking those improvements, I would be happy to help you with changes.

leipert:
I just found #126 and #136.

svanzoest:
I am currently running info an issue with webpack not dealing well with `pkginfo` for `jsonld`

```
test.js:110784
	    throw new Error('Could not find package.json up from ' +
	    ^

Error: Could not find package.json up from 758
    at Function.pkginfo.find (test.js:110784:12)
    at Function.pkginfo.read (test.js:110811:23)
    at module.exports (test.js:110757:22)
    at Object.<anonymous> (test.js:110818:2)
    at Object.<anonymous> (test.js:110824:31)
    at __webpack_require__ (test.js:21:30)
    at wrapper (test.js:109286:4)
    at test.js:109312:4
    at Object.<anonymous> (test.js:109332:4)
    at Object.module.exports.hasExcape (test.js:109334:31)
```

So the work on #126 is appreciated, especially to make it portable to other environments.

davidlehn:
pkginfo usage was removed in https://github.com/digitalbazaar/jsonld.js/commit/17ec94a97a11576d319db643829f720db821d8a1 and released in 0.5.17.

dlongley:
Closing as fixed or now duplicates #399 in some way.

PR #184, opened 2017-05-28
simplify fetching of package version

svanzoest:
This removes the pkginfo dependency to determine the version of the package. 
Using webpack to build my node packages otherwise gives me `package.json` not found.

Addresses #168 and and related to #152

### DIFF
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -8177,9 +8177,7 @@ if(_nodejs) {
   };
 
   // expose version
-  var _module = {exports: {}, filename: __dirname};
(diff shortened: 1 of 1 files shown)


svanzoest:
Tests seem to be failing primarily because of #172

davidlehn:
Fixed in https://github.com/digitalbazaar/jsonld.js/commit/17ec94a97a11576d319db643829f720db821d8a1 and released in 0.5.17.

svanzoest:
👍  Thanks!

PR #185, opened 2017-06-26
Style fixes.

mattcollier:
Keeping these style fixes separate.


.eslintrc
```diff
@@ -0,0 +1,11 @@
+
```
davidlehn: blank line

davidlehn:
Note this should be applied to both 0.4.x and 0.5.x.

PR #186, opened 2017-06-26
Do not promisify `xhr` loader.

mattcollier:
This patch works in e2e tests that perform front-end and back-end jsonld operations.

Depends on: https://github.com/digitalbazaar/bedrock-angular/pull/24

### DIFF
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -289,7 +289,7 @@ jsonld.expand = function(input, options, callback) {
               code: 'loading document failed',
               cause: ex,
               remoteDoc: remoteDoc
-          }));
+            }));
         }
         expand(remoteDoc);
       };
@@ -509,7 +509,7 @@ jsonld.frame = function(input, frame, options, callback) {
               code: 'loading document failed',
               cause: ex,
               remoteDoc: remoteDoc
-          }));
+            }));
         }
         doFrame(remoteDoc);
       };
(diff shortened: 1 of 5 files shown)


js/jsonld.js
```diff
@@ -1938,14 +1944,6 @@ jsonld.documentLoaders.xhr = function(options) {
...
   var headers = buildHeaders(options.headers);
 
-  // use option or, by default, use Promise when its defined
-  var usePromise = ('usePromise' in options ?
-    options.usePromise : (typeof Promise !== 'undefined'));
-  if(usePromise) {
-    return queue.wrapLoader(function(url) {
-      return jsonld.promisify(loader, url);
-    });
-  }
```
dlongley: Needs a similar fix in the other loaders. I also suspect that the `wrapLoader` function needs to change to remove the `_usePromise` flag and treat every loader like it works with both a callback and a Promise (unless the internals of that queue could work with just a callback).

I don't think we can avoid mandating that anyone writing a loader must implement both callbacks and return Promises ... right? We can't automate how custom loaders will work, I don't think, can we? Isn't it too messy and unreliable?

I guess if we want to make it easier for devs to write custom loaders, we should add a helper function that will wrap a callback-only or Promise-only loader function appropriately to expose both mechanisms.

Thoughts?

mattcollier:
@dlongley @davidlehn ready for another pass.

Although phantomjs now has `function.prototype.bind` support, node did add support until v6.  So if we remove the polyfill we would lose support for node <6.

dlongley:
@mattcollier, @davidlehn, let's just roll with this for now and we'll make improvements as we go. We need to get this in before making the major reorganization of the library. We should revisit before a major release post reorg.

davidlehn:
Note this is set to merge to the style branch but the breaking non-style parts should only be pulled into 0.5.x.

dlongley:
I'm closing this because it's been addressed elsewhere.

PR #187, opened 2017-06-27
Remove default ports if found in URLs.

mattcollier:


davidlehn:
Should probably rebase those skip tests commits into one and use a more descriptive log message like "Skip 1.1 tests."  Also would be good to add a line above the regexes saying why they are being skipped since those will go away once the tests suite has better 1.1 test flags.

mattcollier:
@davidlehn @dlongley with these changes, test suite is passing except for promise based documentLoader tests that were not being run due to earlier failures.  Those failures will be addressed by: https://github.com/digitalbazaar/jsonld.js/pull/186

PR #190, opened 2017-06-29
Update jsonld browser + script tag

kimdhamilton:
The jsonld browser + script tag yielded a different canonical form than when I used the library directly. I confirmed there is a later version that yields the expected results. Updating so others don't run into this as well.


mattcollier:
@kimdhamilton Thank you!

davidlehn:
I updated the docs to also use jsonld.min.js and pointed at the cdnjs page to find the latest version.
https://github.com/digitalbazaar/jsonld.js/commit/ddb9e9ccbade55b4d651206fb9898df38d6084ab

PR #193, opened 2017-07-21
Fix _removeDotSegments.

davidlehn:
Fix _removeDotSegments logic to pass all the json-ld.org RFC3986 tests.

Oddly this doesn't require the use of the hasAuthority parameter.  Is this just luck with the current json-ld.org test cases or is there an edge case where this would fail?

This patch requires https://github.com/json-ld/json-ld.org/pull/525 to fully pass the tests.

### DIFF
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -6116,13 +6116,11 @@ function _prependBase(base, iri) {
         var path = base.path;
 
         // append relative path to the end of the last directory from base
-        if(rel.path !== '') {
-          path = path.substr(0, path.lastIndexOf('/') + 1);
-          if(path.length > 0 && path.substr(-1) !== '/') {
-            path += '/';
-          }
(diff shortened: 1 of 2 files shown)


davidlehn:
Will likely hold off on merging this and work on a fix that also handles the current tests that don't remove dot segments from base URIs.

It's also been noted this perhaps should be a minor bump since the outputs could change if these edge case URIs were used as input, even though it's due to a buggy implementation.

davidlehn:
@dlongley Please re-review.  Thanks.

davidlehn:
Note, this now passes all RFC3986 tests as is.

dlongley:
Note: also needs to be ported to 0.5.x if not already.

davidlehn:
Fixed via other commits and released in 0.5.17.

PR #198, opened 2017-08-25
Fix compacting empty list mapped @type: @id

### LINKED ISSUE #189: Handling properties as @container as @list and @type as @id
Hello, recently I created an issue on json-ld.org repo, you can ready about it here: https://github.com/json-ld/json-ld.org/issues/508

Also created PR with test for that which probably will fail for now as fixes are needed:
https://github.com/json-ld/json-ld.org/pull/509

adambro:
It sets default common type to `@id` for an empty list. Extracted from
existing algorithm where `@id` was a default type if list item was not a
value (empty is not a value).

The test case has been marked for 1.1 version of the spec, but has been
considered as an errata to 1.0 spec as well. See comment:
https://github.com/json-ld/json-ld.org/pull/510#issuecomment-312456440

This fixes #189 issue.

### DIFF
--- a/js/jsonld.js
+++ b/js/jsonld.js
@@ -5531,7 +5531,7 @@ function _compactIri(activeCtx, iri, value, relativeTo, reverse) {
       }
       var list = value['@list'];
       var commonLanguage = (list.length === 0) ? defaultLanguage : null;
-      var commonType = null;
+      var commonType = (list.length === 0) ? '@id' : null;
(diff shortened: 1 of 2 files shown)


dlongley:
This was fixed in a more general way by 7fcbdcd088159806b19b5fecdf3d5d7368a1b28f.

adambro:
Thanks for fixing this.

PR #209, opened 2018-01-07
Don't expand non-string native values to node objects.

gkellogg:
As called for in json-ld/json-ld.org#558. There is new text in 1.1 (not dependent on `"@version": 1.1`) to ensure that native types won't be coerced to IRIs. It's quite reasonable for 1.0 processors, as such coercion is nonsensical.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
   feature added complexity and browser issues and the use case is likely
   handled by semantic versioning and using a proper dependency.
 
+## 0.5.x
+- Do not use native types to create IRIs in value expansion.
+
 ## 0.5.15 - 2017-10-16
 
 ### Changed
--- a/lib/expand.js
+++ b/lib/expand.js
@@ -575,11 +575,11 @@ function _expandValue({activeCtx, activeProperty, value}) {
   const type = _getContextValue(activeCtx, activeProperty, '@type');
 
   // do @id expansion (automatic for @graph)
-  if(type === '@id' || (expandedProperty === '@graph' && _isString(value))) {
+  if((type === '@id' || expandedProperty === '@graph') && _isString(value)) {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
   feature added complexity and browser issues and the use case is likely
   handled by semantic versioning and using a proper dependency.
 
+## 0.5.x
+- Do not use native types to create IRIs in value expansion.
+
```
dlongley: We'll want to make this follow our convention. @davidlehn -- any tips/links to point out?

lib/expand.js
```diff
@@ -590,7 +590,7 @@ function _expandValue({activeCtx, activeProperty, value}) {
 
   const rval = {};
 
-  if(type !== null) {
+  if(type && ['@id', '@vocab'].indexOf(type) === -1) {
```
dlongley: Should use `! .includes` instead of `indexOf` check.

davidlehn: With two items is it better to just expand out the two checks?

gkellogg: Perhaps, but it makes it easier to alter in the future; I'll update to use `!.includes`.

gkellogg: Should https://github.com/gkellogg/jsonld.js/blob/no-expand-native-as-object/lib/expand.js#L421-L423 and https://github.com/gkellogg/jsonld.js/blob/no-expand-native-as-object/lib/expand.js#L604  use this convention too?

davidlehn: I think all `indexOf() ===/!== -1` uses can be switched over to `includes()`.

CHANGELOG.md
davidlehn: Added similar comment here: https://github.com/digitalbazaar/jsonld.js/pull/210/files#r160245707
I'll fix it for next 0.5.x release.

PR #210, opened 2018-01-08
Improve conformance of processingMode and converge on array representation of @container

gkellogg:
- Set processingMode from options or first encountered context.
- Use array representation of @container in processing.
- Improved error detection for @container variations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
 
 ### Fixed
 - Do not use native types to create IRIs in value expansion.
+- Improved error detection for @container variations.
+
+### Changed
+- Set processingMode from options or first encountered context.
+- Use array representation of @container in processing.
 
 ## 0.5.15 - 2017-10-16
 
--- a/lib/context.js
+++ b/lib/context.js
@@ -96,7 +96,8 @@ api.process = ({activeCtx, localCtx, options}) => {
           'Unsupported JSON-LD version: ' + ctx['@version'],
           'jsonld.UnsupportedVersion',
           {code: 'invalid @version value', context: ctx});
-      } else if(activeCtx.processingMode && activeCtx.processingMode.indexOf('json-ld-1.1') !== 0) {
+      }
+      if(activeCtx.processingMode && activeCtx.processingMode.indexOf('json-ld-1.1') !== 0) {
         throw new JsonLdError(
           '@version: ' + ctx['@version'] + ' not compatible with ' + activeCtx.processingMode,
           'jsonld.ProcessingModeConflict',
@@ -107,7 +108,7 @@ api.process = ({activeCtx, localCtx, options}) => {
       defined['@version'] = true;
     }
 
-    // If not set explicitly, set processingMode to "json-ld-1.0"
+    // if not set explicitly, set processingMode to "json-ld-1.0"
     rval.processingMode = rval.processingMode || activeCtx.processingMode || 'json-ld-1.0';
 
     // handle @base
@@ -379,7 +380,7 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
   }
 
   if('@container' in value) {
-    // Normalize container to an array form
+    // normalize container to an array form
     const container = _isString(value['@container']) ? [value['@container']] : (value['@container'] || []);
     const validContainers = ['@list', '@set', '@index', '@language'];
     let isValid = true;
@@ -393,14 +394,14 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
       // check container length
       isValid &= container.length <= (hasSet ? 2 : 1);
     } else {
-      // container must be a string which is one of the validContainers
+      // in JSON-LD 1.0, container must not be an array (it must be a string, which is one of the validContainers)
       isValid &= !_isArray(value['@container']);
 
       // check container length
       isValid &= container.length <= 1;
     }
 
-    // Check against valid containers
+    // check against valid containers
     isValid &= container.every(c => validContainers.includes(c));
 
     // @set not allowed with @list
--- a/lib/context.js
+++ b/lib/context.js
@@ -108,7 +108,7 @@ api.process = ({activeCtx, localCtx, options}) => {
     }
 
     // If not set explicitly, set processingMode to "json-ld-1.0"
-    rval.processingMode |= activeCtx.processingMode || 'json-ld-1.0';
+    rval.processingMode = rval.processingMode || activeCtx.processingMode || 'json-ld-1.0';
 
     // handle @base
     if('@base' in ctx) {
--- a/lib/context.js
+++ b/lib/context.js
@@ -108,7 +108,7 @@ api.process = ({activeCtx, localCtx, options}) => {
     }
 
     // If not set explicitly, set processingMode to "json-ld-1.0"
-    rval.processingMode = rval.processingMode || activeCtx.processingMode || 'json-ld-1.0';
+    rval.processingMode |= activeCtx.processingMode || 'json-ld-1.0';
 
     // handle @base
     if('@base' in ctx) {
@@ -391,20 +391,20 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
       validContainers.push('@graph');
 
       // check container length
-      if(container.length > (hasSet ? 2 : 1)) isValid = false;
+      isValid &= container.length <= (hasSet ? 2 : 1);
     } else {
       // container must be a string which is one of the validContainers
-      if(_isArray(value['@container'])) isValid = false;
+      isValid &= !_isArray(value['@container']);
 
       // check container length
-      if(container.length > 1) isValid = false;
+      isValid &= container.length <= 1;
     }
 
     // Check against valid containers
-    if(!container.every(c => validContainers.includes(c))) isValid = false;
+    isValid &= container.every(c => validContainers.includes(c));
 
     // @set not allowed with @list
-    if(hasSet && container.includes('@list')) isValid = false;
+    isValid &= !(hasSet && container.includes('@list'));
 
     if(!isValid) {
       throw new JsonLdError(
--- a/lib/context.js
+++ b/lib/context.js
@@ -96,19 +96,19 @@ api.process = ({activeCtx, localCtx, options}) => {
           'Unsupported JSON-LD version: ' + ctx['@version'],
           'jsonld.UnsupportedVersion',
           {code: 'invalid @version value', context: ctx});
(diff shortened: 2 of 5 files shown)


gkellogg:
@dlongley and @davidlehn This commit fleshes out the support for `@version` and setting `processingMode` based on either API or first encountered context. As a side-effect, it also improved the representation of `@container` and the error checking. You might weigh in on the stylistic choices in `createTermDefinition` to be sure I'm not straying too far from convention. (Ruby has Array intersection, which is a bit cleaner, but this is pretty equivalent).

Most future work, including some graph-container issues, are dependent on this work. As it doesn't really  change behavior for existing uses, it should be safe for the 0.5.x branch.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
   feature added complexity and browser issues and the use case is likely
   handled by semantic versioning and using a proper dependency.
 
+## 0.5.x
+- Set processingMode from options or first encountered context.
+- Use array representation of @container in processing.
+- Improved error detection for @container variations.
+
```
davidlehn: We've been using a mostly http://keepachangelog.com/ format and have sub-headings like Added, Changed, Fixed, etc.  The "Unreleased" part they use was a hassle so I've just been adding new stuff at the top, then on release add the version+date header.  No need to fix this, I'll just update it before releasing next 0.5.x.  No idea how useful the sub-headings are, so could simplify if desired.

gkellogg: Great, thanks.

davidlehn:
There's an issue with the `api.processingMode` usage.  That flag should be per-call vs a flag on the shared API object itself.  This would be an issue if you do two concurrent async calls using 1.0 and 1.1.

It seems to have other side effects too due to perhaps something not being reset (though I think a per-call flag would eliminate this issue)?  If you check out the code in a subdir *without* checking out the test suite, it will run only the other local tests in `tests/`.  The `graph-container.js` ones will fail due to the processing mode being 1.0, despite the input data saying 1.1.  I think this is due to previous tests already setting up 1.0?  If you just run `JSONLD_TESTS=`pwd`/tests/graph-container.js npm t` it would work fine.  I'm not sure why the test-suite plus the extra tests works.

gkellogg:
> I'm not sure why the test-suite plus the extra tests works.
This is probably an artifact of using the API object to hold this; it's created once, and re-used for each test, apparently.

I'll look deeper to see what the issue of keeping it on the active context is, if any.

lib/context.js
```diff
@@ -370,31 +379,42 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
...
+    let isValid = true;
+    let hasSet = container.includes('@set');
+
+    // JSON-LD 1.1 support
+    if(activeCtx.processingMode && activeCtx.processingMode.indexOf('json-ld-1.1') === 0) {
+      // TODO: @id and @type
+      validContainers.push('@graph');
+
+      // check container length
+      if(container.length > (hasSet ? 2 : 1)) isValid = false;
```
davidlehn: The style we've been using for `if`s is, in general, to be multiline with `{}`s.  A bit more verbose but is consistent.  Could also use a `isValid &= (...)` style for these.
@dlongley comments?

gkellogg: Just need to reverse the logic sense, but it is more concise.

lib/context.js
```diff
@@ -96,11 +96,20 @@ api.process = ({activeCtx, localCtx, options}) => {
           'Unsupported JSON-LD version: ' + ctx['@version'],
           'jsonld.UnsupportedVersion',
           {code: 'invalid @version value', context: ctx});
+      } else if(activeCtx.processingMode && activeCtx.processingMode.indexOf('json-ld-1.1') !== 0) {
```
dlongley: The above `if` throws an error, so our style is not to use an `else if` but to just use a regular `if` on a new line here, i.e. if an `if` exits the function via throwing or returning, there's no need to use `else`, just use `if`.

lib/context.js
```diff
@@ -96,11 +96,20 @@ api.process = ({activeCtx, localCtx, options}) => {
...
+          '@version: ' + ctx['@version'] + ' not compatible with ' + activeCtx.processingMode,
+          'jsonld.ProcessingModeConflict',
+          {code: 'processing mode conflict', context: ctx});
       }
+      rval.processingMode = 'json-ld-1.1';
       rval['@version'] = ctx['@version'];
       defined['@version'] = true;
     }
 
+    // If not set explicitly, set processingMode to "json-ld-1.0"
```
dlongley: Very minor nit, but we typically don't capitalize at the start of a `//` comment. Comments that require lengthy explanations and full sentences get the `/* */` treatment. This is a short comment so here we'd use `//` and not capitalize, i.e.:

```js
// if not set explicitly, set processingMode to "json-ld-1.0"
```

lib/context.js
```diff
@@ -370,31 +379,42 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
...
+
+    // JSON-LD 1.1 support
+    if(activeCtx.processingMode && activeCtx.processingMode.indexOf('json-ld-1.1') === 0) {
+      // TODO: @id and @type
+      validContainers.push('@graph');
+
+      // check container length
+      isValid &= container.length <= (hasSet ? 2 : 1);
+    } else {
+      // container must be a string which is one of the validContainers
```
dlongley: May want this comment to be a bit more explicit since the check doesn't directly involve strings:

```js
// container must not be an array (it must be a string, which is one of the validContainers)
```

lib/context.js
```diff
@@ -370,31 +379,42 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
...
-      if(!isValid) {
-        throw new JsonLdError(
-          'Invalid JSON-LD syntax; @context @container value must be ' +
-          'one of the following: ' + validContainers.join(', '),
-          'jsonld.SyntaxError',
-          {code: 'invalid container mapping', context: localCtx});
-      }
+    // Normalize container to an array form
+    const container = _isString(value['@container']) ? [value['@container']] : (value['@container'] || []);
+    const validContainers = ['@list', '@set', '@index', '@language'];
```
dlongley: Should probably use a `Set` for `validContainers` instead of an array -- and then use `.has` instead of `.includes`. But just a minor nit. Just pointing out that we should make better use of the newer ES6/ES7/whatever features.

gkellogg: Looks like node 6 BUNDLER=webpack build doesn't like the use of `Set`: https://travis-ci.org/gkellogg/jsonld.js/jobs/326948773. Is there some transformation that is supposed to happen automatically?

davidlehn: babel handles syntax, and apis come from core-js.  In theory, you would just add `core-js/fn/set` to `webpack.config.js` and `tests/test-karma.js` along with the other core-js things.  But I just tried and karma tests still fail even though looks like Set code is in the built code.

gkellogg: I'm going to back out the change to use `Set`, which could probably be used in other places anywhere, and it's really the point of this branch.

davidlehn: Ok... the fix I suggested mostly works.  It's just a few error tests that fail for some reason and only in PhantomJS.

lib/context.js
```diff
@@ -370,31 +380,42 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
...
-      if(!isValid) {
-        throw new JsonLdError(
-          'Invalid JSON-LD syntax; @context @container value must be ' +
-          'one of the following: ' + validContainers.join(', '),
-          'jsonld.SyntaxError',
-          {code: 'invalid container mapping', context: localCtx});
-      }
+    // normalize container to an array form
+    const container = _isString(value['@container']) ? [value['@container']] : (value['@container'] || []);
+    const validContainers = ['@list', '@set', '@index', '@language'];
```
davidlehn: This list is just static for 1.0 and 1.1 right?  May be more efficient to hoist it out to two static lists and choose which one to use based on mode at runtime.  Would avoid reallocation and updating per-call.

tests/test-common.js
```diff
@@ -42,7 +42,7 @@ const TEST_TYPES = {
   },
   'jld:ExpandTest': {
     skip: {
-      specVersion: ['json-ld-1.1']
+      regex: [/#t[cmn]/, /#t008./]
```
davidlehn: Not sure how best to handle this, but I had to change it to `/#t008[0-7]/` so it ran 0088 and the new tests from https://github.com/json-ld/json-ld.org/pull/567.

lib/context.js
gkellogg: Yes, that's what I do with my implementation, but I was just trying to follow the existing pattern. I can update to make these static arrays.

tests/test-common.js
gkellogg: That certainly will do the job.

lib/context.js
gkellogg: Looking through the code, there's little uniformity for how such things are handled. Elsewhere in expand.js, there are inline arrays of keywords, and `api.isKeyword` is defined to use a switch statement to detect. Creating a top-level static array seems like yet a third way. Is a `switch` statement faster than `Array.includes`? Arguably, `Set.has` would be fast, if it worked consistently in `test-karma`. One could argue for a large refactor to unify the way all such lists are set and maintained.

Given that the `validContainers` is code I inherited, I'd rather not keep creeping this PR with things that should be done in a more general refactor.

lib/context.js
```diff
@@ -379,37 +380,37 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
...
+    isValid &= container.every(c => validContainers.has(c));
 
     // @set not allowed with @list
     isValid &= !(hasSet && container.includes('@list'));
 
     if(!isValid) {
       throw new JsonLdError(
         'Invalid JSON-LD syntax; @context @container value must be ' +
-        'one of the following: ' + validContainers.join(', '),
+        'one of the following: ' + Array.from(validContainers).join(', '),
```
dlongley: Can use spread operator instead of `Array.from`:

`[...validContainers].join(', '),`

lib/context.js
dlongley: I agree that a larger refactor to make this stuff more consistent should be done in a separate PR.

dlongley:
One more minor nit -- and fixups due to conflicts and I'll approve.

gkellogg:
@dlongley I don't see any conflicts to resolve, can you be more specific? What is the other minor nit that needs to be addressed?

dlongley:
@gkellogg, the nit no longer applies since you reverted the `Set` change (I suggested using the spread operator instead of `Array.from`) until we can do a larger refactor.

The conflict is listed here in the test report. It needs to be rebased and conflicts resolved and then force pushed.

Issue #207, opened 2017-11-10
Handling of expansion of empty `@base`.


davidlehn:
Current code, after RFC3986 fixes, handles empty `@base` in an unexpected way.  See example in https://github.com/json-ld/json-ld.org/issues/535.

- Determine if undefined empty `@base` behavior is ok.
- Add json-ld.org test suite test for behavior one way or another?
- Add local test so code doesn't change the "undefined" behavior?
- Raise error and suggest using `null`?
- Update lib code that is setting base to "" if something else (null?) is more appropriate.

davidlehn:
Looks like empty and relative values are allowed for `@base`.  Code needs to be updated and hopefully the main test suite will grow tests for this.

davidlehn:
Fixed and test suite tests added.  Released in 0.5.17.

PR #211, opened 2018-01-10
Handle empty and relative @base.

davidlehn:
This handles the empty and relative `@base` tests from https://github.com/json-ld/json-ld.org/pull/567.

The absolute and relative iri detection is very weak.  Probably best to have a future pedantic mode that does slow checks on everything and just assume reasonable input for now.  I left a comment pointing to regexes.

Also seems the `parseUrl()` part isn't needed here for abs or rel iris?

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -16,6 +16,7 @@ const {
 
 const {
   isAbsolute: _isAbsoluteIri,
+  isRelative: _isRelativeIri,
   prependBase,
   parse: parseUrl
 } = require('./url');
@@ -115,24 +116,19 @@ api.process = ({activeCtx, localCtx, options}) => {
     if('@base' in ctx) {
       let base = ctx['@base'];
 
-      // clear base
       if(base === null) {
-        base = null;
-      } else if(!_isString(base)) {
+        // no action
+      } else if(_isAbsoluteIri(base)) {
+        base = parseUrl(base);
+      } else if(_isRelativeIri(base)) {
+        base = parseUrl(prependBase(activeCtx['@base'].href, base));
+      } else {
         throw new JsonLdError(
           'Invalid JSON-LD syntax; the value of "@base" in a ' +
-          '@context must be a string or null.',
-          'jsonld.SyntaxError', {code: 'invalid base IRI', context: ctx});
-      } else if(base !== '' && !_isAbsoluteIri(base)) {
-        throw new JsonLdError(
-          'Invalid JSON-LD syntax; the value of "@base" in a ' +
-          '@context must be an absolute IRI or the empty string.',
+          '@context must be an absolute IRI, a relative IRI, or null.',
           'jsonld.SyntaxError', {code: 'invalid base IRI', context: ctx});
       }
 
-      if(base !== null) {
-        base = parseUrl(base || '');
-      }
       rval['@base'] = base;
       defined['@base'] = true;
     }
--- a/lib/url.js
+++ b/lib/url.js
@@ -269,12 +269,13 @@ api.removeDotSegments = path => {
 // TODO: time better isAbsolute/isRelative checks using full regexes:
 // http://jmrware.com/articles/2009/uri_regexp/URI_regex.html
 
-// regex to check for starting scheme and ':'
-const isAbsoluteRegex = /^[A-Za-z][A-Za-z0-9+-.]*:/
+// regex to check for absolute IRI (starting scheme and ':') or blank node IRI
+const isAbsoluteRegex = /^([A-Za-z][A-Za-z0-9+-.]*|_):/;
 
 /**
- * Returns true if the given value is an absolute IRI, false if not.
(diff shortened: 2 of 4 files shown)


davidlehn:
https://github.com/digitalbazaar/jsonld.js/issues/207

lib/context.js
```diff
@@ -105,24 +106,19 @@ api.process = ({activeCtx, localCtx, options}) => {
...
-      // clear base
       if(base === null) {
-        base = null;
-      } else if(!_isString(base)) {
-        throw new JsonLdError(
-          'Invalid JSON-LD syntax; the value of "@base" in a ' +
-          '@context must be a string or null.',
-          'jsonld.SyntaxError', {code: 'invalid base IRI', context: ctx});
-      } else if(base !== '' && !_isAbsoluteIri(base)) {
+        // no action
```
gkellogg: Testing this would seem to do nothing for any relative (non-absolute) IRI which isn't the empty string, and not execute the code at 114 below.

lib/context.js
```diff
@@ -105,24 +106,19 @@ api.process = ({activeCtx, localCtx, options}) => {
...
+      } else if(_isAbsoluteIri(base)) {
+        base = parseUrl(base);
+      } else if(_isRelativeIri(base)) {
+        base = parseUrl(prependBase(activeCtx['@base'].href, base));
+      } else {
```
gkellogg: This seems right. One thing I tried on my implementation was the following:

```json
{
  "@context": [{
    "@base": "http://foo.bar/./baz/"
  }, {
    "@base": "example/"
  }],
  "@id": "foobar",
  "http:example.org/": "v"
}
```

This should result in:

```json
[
  {
    "@id": "http://foo.bar/baz/example/foobar",
    "http:example.org/": [
      {
        "@value": "v"
      }
    ]
  }
]
```

lib/url.js
```diff
@@ -266,11 +266,25 @@ api.removeDotSegments = path => {
...
 /**
  * Returns true if the given value is an absolute IRI, false if not.
+ * Note: this is a weak check.
  *
  * @param v the value to check.
  *
  * @return true if the value is an absolute IRI, false if not.
  */
-api.isAbsolute = v => types.isString(v) && v.indexOf(':') !== -1;
+api.isAbsolute = v => types.isString(v) && v.includes(':');
```
gkellogg: To do this right, along with other URI processing, really requires full parsing based on RFC3986, but if it passes tests, you probably don't need to go that far. I have a full implementation in Ruby [here](https://github.com/ruby-rdf/rdf/blob/develop/lib/rdf/model/uri.rb), which is unencumbered. You might think that rdflib.js would too, but it seems light.

davidlehn: This and the new isRelative() call are obviously not very robust.  We could (should?) easily craft tests that break the code.  Like make a relative IRI with a ':' in it.  I did add a comment above for a page with regexes.  Was trying to avoid adding too much to this PR.

gkellogg: An absolute URI has a schema followed by a colon, so something like `/^\w+:/` might work.

davidlehn: I changed it check via a proper scheme regex.

lib/context.js
davidlehn: I added this example to test suite.

lib/context.js
gkellogg: @davidlehn common on this?

davidlehn: I don't understand the issue.  It does nothing for null, else do something if absolute, else do something if relative (which is a lame check so will just run on any string), else fail (ie, non string or something).

gkellogg: This would seem to be true of `base = 'a/b'`, as `base !== '' && !_isAbsoluteIri(base)`. This would seem to be where 114 comes in `base = parseUrl(prependBase(activeCtx['@base'].href, base))`.

davidlehn: I'm still confused.  Is that line 114 wrong?  I guessed what to do there.  The code previously didn't support relative IRIs at all.  What's an example that would fail?

gkellogg: No, you're right, I was getting confused by old vs new code. This should be good.

lib/context.js
dlongley: In this example, shouldn't the second context `@base` simply override the first's `@base`? The combination of them in the output seemed surprising to me. What am I missing?

davidlehn: It's building the base as it processes things.  You think the secondary one here should override the doc base instead?  So the id would be "{docbase}/example/foobar"?  I could see it going either way.  What crazy person would be doing this anyway?

dlongley: That's how every other `@context` rule works, from what I remember. There's no such thing as partially or progressively updating a definition in a `@context`, rather, you replace it entirely.

gkellogg: It seems to me that @base is always resolved against the current base.

PR #212, opened 2018-01-12
Graph id type maps

gkellogg:
This adds support for graph indexing on `@id` and `@index`, as well as id maps and type maps.

It is based on the `processingMode` branch, but is targeted for 0.6.x. The 0.6.x branch will require rebasing from 0.5.x until that branch is completed.

### DIFF
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -330,8 +330,8 @@ api.compact = ({
 
         // Graph object compaction cases
         if(isGraph) {
-          if(container.includes('@graph') && container.includes('@id')) {
-            // container includes @graph and @id
+          if(container.includes('@graph') &&
+            (container.includes('@id') || container.includes('@index') && _isSimpleGraph(expandedItem))) {
             // get or create the map object
             let mapObject;
             if(itemActiveProperty in rval) {
@@ -339,22 +339,14 @@ api.compact = ({
             } else {
               rval[itemActiveProperty] = mapObject = {};
             }
+
+            // TODO: future change will use @none in both cases.
+            const key = container.includes('@id') ?
+              (expandedItem['@id'] || issuer.getId(null)) :
+              (expandedItem['@index'] || '@none');
             // add compactedItem to map, using value of `@id` or a new blank node identifier
             _addValue(
-              mapObject, (expandedItem['@id'] || issuer.getId(null)), compactedItem,
-              {propertyIsArray: (!options.compactArrays || container.includes('@set'))});
-          } else if(container.includes('@graph') && container.includes('@index') && _isSimpleGraph(expandedItem)) {
-            // container includes @graph and @index and item is a simple graph
-            // get or create the map object
-            let mapObject;
-            if(itemActiveProperty in rval) {
-              mapObject = rval[itemActiveProperty];
-            } else {
-              rval[itemActiveProperty] = mapObject = {};
-            }
-            // add compactedItem to map, using value of `@index` or `@none`
-            _addValue(
-              mapObject, (expandedItem['@index'] || '@none'), compactedItem,
+              mapObject, key, compactedItem,
               {propertyIsArray: (!options.compactArrays || container.includes('@set'))});
           } else if(container.includes('@graph') && _isSimpleGraph(expandedItem)) {
             // container includes @graph but not @id or @index and value is a simple graph object
@@ -387,13 +379,14 @@ api.compact = ({
           container.includes('@id') || container.includes('@type')) {
           // handle language and index maps
           // get or create the map object
-          let mapObject, key;
+          let mapObject;
           if(itemActiveProperty in rval) {
             mapObject = rval[itemActiveProperty];
           } else {
             rval[itemActiveProperty] = mapObject = {};
           }
 
+          let key;
           if(container.includes('@language')) {
           // if container is a language map, simplify compacted value to
           // a simple string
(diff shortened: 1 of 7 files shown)


lib/graphTypes.js
```diff
@@ -67,6 +67,21 @@ api.isList = v =>
...
+ * @return true if the value is a @graph, false if not.
+ */
+api.isGraph = v => {
+  // Note: A value is a graph if all of these hold true:
+  // 1. It is an object.
+  // 2. It has an `@graph` key.
+  // 3. It may have '@id' or '@index'
+  return types.isObject(v) &&
+    '@graph' in v &&
+    Object.keys(v).filter(key => key != '@id' && key != '@index').length === 1;
```
davidlehn: Use `!==`.

lib/expand.js
```diff
@@ -390,7 +418,11 @@ api.expand = ({
     }
 
     // convert expanded value to @graph if container specifies it
-    if(container.includes('@graph')) {
+    // and value is not, itself, a graph
+    // index cases handled above
+    if(container.includes('@graph') &&
+      !container.some(key => key == '@id' || key == '@index') &&
```
davidlehn: Use `===`.

lib/compact.js
```diff
@@ -308,40 +313,100 @@ api.compact = ({
...
+            let mapObject;
+            if(itemActiveProperty in rval) {
+              mapObject = rval[itemActiveProperty];
+            } else {
+              rval[itemActiveProperty] = mapObject = {};
+            }
+            // add compactedItem to map, using value of `@index` or `@none`
+            _addValue(
+              mapObject, (expandedItem['@index'] || '@none'), compactedItem,
+              {propertyIsArray: (!options.compactArrays || container.includes('@set'))});
```
dlongley: The code in these first two cases is really similar and could potentially be combined -- but that can wait until we do some clean up later. Just making a note of it.

lib/compact.js
```diff
@@ -308,40 +313,100 @@ api.compact = ({
...
+            _addValue(
+              rval, itemActiveProperty, compactedItem,
+              {propertyIsArray: (!options.compactArrays || container.includes('@set'))});
+          }
+        } else if(container.includes('@language') || container.includes('@index') ||
+          container.includes('@id') || container.includes('@type')) {
+          // handle language and index maps
           // get or create the map object
-          let mapObject;
+          let mapObject, key;
```
dlongley: Declare vars on new lines:

```js
let mapObject;
let key;
```

But the `key` declaration should really be moved down to where it's immediately relevant, so right above its first use.

lib/compact.js
```diff
@@ -308,40 +313,100 @@ api.compact = ({
...
+            key = expandedItem['@language'];
+          } else if(container.includes('@index')) {
+            key = expandedItem['@index'];
+          } else if(container.includes('@id')) {
+            const idKey = api.compactIri({activeCtx, iri: '@id', vocab: true});
+            key = compactedItem[idKey] || '@none';
+            delete compactedItem[idKey];
+          } else if(container.includes('@type')) {
+            const typeKey = api.compactIri({activeCtx, iri: '@type', vocab: true});
+            var types;
```
dlongley: Use `let types`, not `var`.

lib/compact.js
```diff
@@ -308,40 +313,100 @@ api.compact = ({
...
+            [key, ...types] = [].concat(compactedItem[typeKey] || ['@none']);
+            switch(types.length) {
+            case 0: delete compactedItem[typeKey]; break;
+            case 1: compactedItem[typeKey] = types[0]; break;
+            default: compactedItem[typeKey] = types; break;
```
dlongley: Use line breaks:

```js
switch(types.length) {
case 0:
  delete compactedItem[typeKey];
  break;
case 1:
  compactedItem[typeKey] = types[0];
  break;
default:
  compactedItem[typeKey] = types;
  break;
}
```

Or shorten to if/else:

```js
if(types.length === 0) {
  delete compactedItem[typeKey];
} else if(types.length === 1) {
  compactedItem[typeKey] = types[0];
} else {
  compactedItem[typeKey] = types;
}
```

lib/compact.js
```diff
@@ -308,40 +313,100 @@ api.compact = ({
...
+            }
           }
 
           // add compact value to map object using key from expanded value
           // based on the container type
-          const c = container.includes('@language') ? '@language' : '@index';
-          _addValue(mapObject, expandedItem[c], compactedItem);
+          _addValue(mapObject, key, compactedItem,
```
dlongley: Our coding style rule for long parameter lists is:

When moving any function parameters down a line, move them all down a line (and then add new lines as needed to fit within 80-chars, but do not leave any on the same line as the function as it improves readability):

```js
_addValue(
  mapObject, key, compactedItem, {propertyIsArray: container.includes('@set')});
```

lib/context.js
```diff
@@ -389,10 +389,28 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
...
-      isValid &= container.length <= (hasSet ? 2 : 1);
+      if(container.includes('@list')) {
+        if(container.length !== 1) {
+          throw new JsonLdError(
+            'Invalid JSON-LD syntax; @context @container with @list must have no other values',
+            'jsonld.SyntaxError',
+            {code: 'invalid container mapping', context: localCtx});
+        }
+      } else if(container.includes('@graph')) {
+        if(container.some(key => key != '@graph' && key != '@id' && key != '@index' && key != '@set')) {
```
dlongley: Use triple equals.

gkellogg:
I made the requested changes. Note that this will need a rebase to fit with the other two PRs.

lib/compact.js
```diff
@@ -411,11 +475,19 @@ api.compactIri = ({
...
+      containers.push('@graph@index@set');
+      containers.push('@graph@index');
+      containers.push('@graph@id@set');
+      containers.push('@graph@id');
       containers.push('@graph@set');
       containers.push('@graph');
+    } else if(_isObject(value) && !_isValue(value)) {
+      containers.push('@id@set');
+      containers.push('@id');
+      containers.push('@type@set');
```
davidlehn: Flipped order?

gkellogg: Should be `@set@type`, and can `push` can take multiple arguments.

Based on https://github.com/json-ld/json-ld.org/pull/569 it's even more complicated, if we want to prioritize order based on the actual value, and allow `@none` to be used. The work's not all done fo this yet, but may as well get this in order.

lib/compact.js
gkellogg: This better?

dlongley: I meant the code is the same for the cases where `@index` and where `@id` are present other than the second parameter to `_addValue`.

PR #213, opened 2018-01-13
Scoped contexts

gkellogg:
Adds support for scoped contexts on properties and types for compaction and expansion.

### DIFF
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -21,7 +21,8 @@ const {
 const {
   expandIri: _expandIri,
   getContextValue: _getContextValue,
-  isKeyword: _isKeyword
+  isKeyword: _isKeyword,
+  process: _processContext
 } = require('./context');
 
 const {
@@ -95,6 +96,12 @@ api.compact = ({
     return rval;
   }
 
+  // use any scoped context on activeProperty
+  const ctx = _getContextValue(activeCtx, activeProperty, '@context');
+  if(ctx) {
+    activeCtx = _processContext({activeCtx, localCtx: ctx, options});
+  }
+
   // recursively compact object
   if(_isObject(element)) {
     if(options.link && '@id' in element && element['@id'] in options.link) {
@@ -154,8 +161,16 @@ api.compact = ({
           // expanded value must be a @type array
           compactedValue = [];
           for(let vi = 0; vi < expandedValue.length; ++vi) {
-            compactedValue.push(api.compactIri(
-              {activeCtx, iri: expandedValue[vi], relativeTo: {vocab: true}}));
+            const compactedType = api.compactIri(
+              {activeCtx, iri: expandedValue[vi], relativeTo: {vocab: true}})
+
+            // Use any scoped context defined on this value
+              const ctx = _getContextValue(activeCtx, compactedType, '@context');
+              if(ctx) {
+                activeCtx = _processContext({activeCtx, localCtx: ctx, options});
+              }
+
+            compactedValue.push(compactedType);
           }
         }
 
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -165,10 +165,10 @@ api.compact = ({
               {activeCtx, iri: expandedValue[vi], relativeTo: {vocab: true}})
 
             // Use any scoped context defined on this value
-              const ctx = _getContextValue(activeCtx, compactedType, '@context');
-              if(ctx) {
-                activeCtx = _processContext({activeCtx, localCtx: ctx, options});
-              }
+            const ctx = _getContextValue(activeCtx, compactedType, '@context');
+            if(ctx) {
+              activeCtx = _processContext({activeCtx, localCtx: ctx, options});
+            }
 
             compactedValue.push(compactedType);
           }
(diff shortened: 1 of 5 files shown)


lib/context.js
```diff
@@ -979,6 +990,13 @@ function _findContextUrls(input, urls, replace, base) {
             // @context URL found
             urls[_ctx] = false;
           }
+        } else {
+          // look for scoped context
```
gkellogg: @dlongley I think this handles retrieving scoped contexts properly when scanning for contexts, but please consider.

lib/compact.js
```diff
@@ -154,8 +161,16 @@ api.compact = ({
...
-            compactedValue.push(api.compactIri(
-              {activeCtx, iri: expandedValue[vi], relativeTo: {vocab: true}}));
+            const compactedType = api.compactIri(
+              {activeCtx, iri: expandedValue[vi], relativeTo: {vocab: true}})
+
+            // Use any scoped context defined on this value
+              const ctx = _getContextValue(activeCtx, compactedType, '@context');
+              if(ctx) {
+                activeCtx = _processContext({activeCtx, localCtx: ctx, options});
+              }
```
davidlehn: Too much indent.

lib/context.js
dlongley: Heh, I think tests would consider this better than a human. :)

lib/context.js
```diff
@@ -426,6 +426,17 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
     mapping['@container'] = container;
   }
 
+  // scoped contexts
+  if('@context' in value) {
+    if(activeCtx.processingMode && activeCtx.processingMode.indexOf('json-ld-1.1') !== 0) {
```
dlongley: If we're going to have a decent number of these checks, we should probably make a helper function that takes something like the active context, the processing mode required, and optional error text+details to use if the check fails.

lib/expand.js
```diff
@@ -143,12 +143,31 @@ api.expand = ({
       {activeCtx, localCtx: element['@context'], options});
   }
 
+  let keys = Object.keys(element).sort();
+
+  // look for scoped context on @type
+  for(let ki = 0; ki < keys.length; ++ki) {
```
dlongley: If you don't need the `ki` index var, use `for...of` and declare your `const key` within it:

```js
for(const key of keys) {
   // ...
```

lib/expand.js
```diff
@@ -143,12 +143,31 @@ api.expand = ({
...
+    const key = keys[ki];
+    let expandedProperty = _expandIri(activeCtx, key, {vocab: true});
+    if(expandedProperty === '@type') {
+
```
dlongley: Remove empty whitespace line.

lib/expand.js
```diff
@@ -143,12 +143,31 @@ api.expand = ({
       {activeCtx, localCtx: element['@context'], options});
   }
 
+  let keys = Object.keys(element).sort();
+
+  // look for scoped context on @type
+  for(let ki = 0; ki < keys.length; ++ki) {
+    const key = keys[ki];
+    let expandedProperty = _expandIri(activeCtx, key, {vocab: true});
```
dlongley: Use `const` if `expandedProperty` is not changed later.

lib/expand.js
```diff
@@ -143,12 +143,31 @@ api.expand = ({
...
+    if(expandedProperty === '@type') {
+
+      // set scopped contexts from @type
+      const types = [].concat(element[key]);
+      for(let ii = 0; ii < types.length; ++ii) {
```
dlongley: Use `for(const type of types)`

lib/expand.js
```diff
@@ -143,12 +143,31 @@ api.expand = ({
...
+    }
+  }
+
   // expand the active property
   const expandedActiveProperty = _expandIri(
     activeCtx, activeProperty, {vocab: true});
 
   let rval = {};
-  let keys = Object.keys(element).sort();
   for(let ki = 0; ki < keys.length; ++ki) {
```
dlongley: Use `for(const key of keys)`

lib/expand.js
```diff
@@ -313,15 +332,23 @@ api.expand = ({
       continue;
     }
 
-    const container = _getContextValue(activeCtx, key, '@container') || [];
+    // use potential scoped context for key
+    let termCtx = activeCtx;
+    const ctx = _getContextValue(activeCtx, key, '@context');
+    if(ctx) {
+      termCtx = _processContext(
+        {activeCtx, localCtx: ctx, options});
```
dlongley: Seems like this param would fit on the same line as `_processContext` (i.e. no need to drop it down).

lib/context.js
gkellogg: The processingMode check can be replaced with the separate check-td-keys, which will simplify these tests, but a helper function is probably worthwhile.

gkellogg:
I've made the requested changes, and rebased onto check-td-keys. You'll want to be sure that branch is pulled to 0.5.x before this is pulled to 0.6.x.

davidlehn:
Side note: There's a bunch of code here and elsewhere that hasn't changed to for..of or is using let vs const.  I had skipped commenting on that figuring we'd address it in later updates.

gkellogg:
@davidlehn I'd rather not pull this into 0.6.x before #214 is pulled into 0.5, as it depends on the functionality there.

PR #217, opened 2018-01-21
Support nested property expansion and compaction.

gkellogg:
Note, `_expandObject` is mostly a simple re-factor of the body of `expand`.

Processing of `nests` is done at the end, rather than when encountered, which is required to produce results expected by the test suite.

### DIFF
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -143,7 +143,7 @@ api.compact = ({
 
     // process element keys in order
     const keys = Object.keys(element).sort();
-    for(let expandedProperty of keys) {
+    for(const expandedProperty of keys) {
       const expandedValue = element[expandedProperty];
 
       // compact @id and @type(s)
@@ -160,7 +160,7 @@ api.compact = ({
         } else {
           // expanded value must be a @type array
           compactedValue = [];
-          for(let expandedIri of expandedValue) {
+          for(const expandedIri of expandedValue) {
             const compactedType = api.compactIri(
               {activeCtx, iri: expandedIri, relativeTo: {vocab: true}})
 
@@ -193,7 +193,7 @@ api.compact = ({
         });
 
         // handle double-reversed properties
-        for(let compactedProperty in compactedValue) {
+        for(const compactedProperty in compactedValue) {
           if(activeCtx.mappings[compactedProperty] &&
             activeCtx.mappings[compactedProperty].reverse) {
             const value = compactedValue[compactedProperty];
@@ -270,7 +270,7 @@ api.compact = ({
       }
 
       // recusively process array values
-      for(let expandedItem of expandedValue) {
+      for(const expandedItem of expandedValue) {
         // compact property and get container type
         const itemActiveProperty = api.compactIri({
           activeCtx,
@@ -844,7 +844,7 @@ api.removePreserve = (ctx, input, options) => {
     }
 
     // recurse through properties
-    for(let prop in input) {
+    for(const prop in input) {
       let result = api.removePreserve(ctx, input[prop], options);
       const container = _getContextValue(ctx, prop, '@container') || [];
       if(options.compactArrays && _isArray(result) && result.length === 1 &&
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -143,8 +143,7 @@ api.compact = ({
 
     // process element keys in order
     const keys = Object.keys(element).sort();
-    for(let ki = 0; ki < keys.length; ++ki) {
-      const expandedProperty = keys[ki];
+    for(let expandedProperty of keys) {
       const expandedValue = element[expandedProperty];
 
       // compact @id and @type(s)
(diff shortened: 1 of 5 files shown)


lib/compact.js
```diff
@@ -161,9 +160,9 @@ api.compact = ({
         } else {
           // expanded value must be a @type array
           compactedValue = [];
-          for(let vi = 0; vi < expandedValue.length; ++vi) {
+          for(let expandedIri of expandedValue) {
```
dlongley: Looks like could use `const expandedIri` instead of `let`.

lib/compact.js
```diff
@@ -257,14 +256,21 @@ api.compact = ({
...
         _addValue(
-          rval, itemActiveProperty, expandedValue, {propertyIsArray: true});
+          nestResult, itemActiveProperty, expandedValue, {propertyIsArray: true});
       }
 
       // recusively process array values
-      for(let vi = 0; vi < expandedValue.length; ++vi) {
-        const expandedItem = expandedValue[vi];
-
+      for(let expandedItem of expandedValue) {
```
dlongley: Use `const` if `expandedItem` is not set anywhere in the loop.

lib/compact.js
```diff
@@ -143,8 +143,7 @@ api.compact = ({
 
     // process element keys in order
     const keys = Object.keys(element).sort();
-    for(let ki = 0; ki < keys.length; ++ki) {
-      const expandedProperty = keys[ki];
+    for(let expandedProperty of keys) {
```
dlongley: Use `const` if `expandedProperty` is not set anywhere in the loop.

lib/compact.js
```diff
@@ -906,3 +924,18 @@ function _selectTerm(
...
+ * @param activeCtx the active context.
+ * @param nestProperty a term in the active context or `@nest`.
+ */
+function _checkNestProperty(activeCtx, nestProperty) {
+  if(_expandIri(activeCtx, nestProperty, {vocab: true}) !== '@nest') {
+    throw new JsonLdError(
+      'JSON-LD compact error; nested property must have an @nest value resolving to @nest.',
+      'jsonld.SyntaxError', {code: 'invalid @nest value'});
+  }
+}
```
dlongley: Add EOL

lib/context.js
```diff
@@ -478,6 +484,17 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
     mapping['@language'] = language;
   }
 
+  if('@nest' in value) {
+    let nest = value['@nest'];
```
dlongley: `const nest`

lib/expand.js
```diff
@@ -145,9 +145,8 @@ api.expand = ({
       {activeCtx, localCtx: element['@context'], options});
   }
 
-  let keys = Object.keys(element).sort();
-
   // look for scoped context on @type
+  let keys = Object.keys(element).sort();
```
dlongley: `const keys`

lib/expand.js
```diff
@@ -162,11 +161,172 @@ api.expand = ({
     }
   }
 
+
```
dlongley: Unnecessary added whitespace

lib/expand.js
```diff
@@ -475,130 +640,32 @@ api.expand = ({
...
-        insideList
-      });
-      if(mapped !== undefined) {
-        rval = mapped;
-      } else {
-        rval = null;
+  // expand each nested key
+  for(let key of nests) {
+    const nestedValues = _isArray(element[key]) ? element[key] : [element[key]];
+    for(let nv of nestedValues) {
```
dlongley: `const`?

lib/expand.js
```diff
@@ -475,130 +640,32 @@ api.expand = ({
...
-        element,
-        options,
-        insideList
-      });
-      if(mapped !== undefined) {
-        rval = mapped;
-      } else {
-        rval = null;
+  // expand each nested key
+  for(let key of nests) {
```
dlongley: `const`?

gkellogg:
Those two are actually variables, as they're modified inline. This is questionable style, but for whatever reason, I chose not to create new variable to hold a possibly modified version.

dlongley:
@gkellogg,

I see that the `keys` one is modified, but the `nest` one is not, AFAICT:

https://github.com/digitalbazaar/jsonld.js/blob/7baabf86f840c92091e23eacdef2513c1c08932d/lib/context.js#L487-L496

dlongley:
@gkellogg, also please make sure to use rebase rather than merge to bring in other updates from 0.6.x, it keeps the history cleaner.

gkellogg:
I've been re-basing, but the GitHub UI for resolving conflicts seems to have done a merge, I can probably re-do on the desktop.

gkellogg:
Force pushed rebase.

PR #218, opened 2018-01-21
Remove shortcut from compactIri when iri is a keyword

gkellogg:
which was preventing cases when keyword alias had `@type` defined, as in Compact 0073.

### DIFF
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -393,7 +393,7 @@ api.compactIri = ({
 
   const inverseCtx = activeCtx.getInverse();
 
-  // if term is a keyword, it can may be compacted to a simple alias
+  // if term is a keyword, it may be compacted to a simple alias
   if(_isKeyword(iri) &&
     iri in inverseCtx &&
     '@none' in inverseCtx[iri] &&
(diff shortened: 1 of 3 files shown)


lib/compact.js
```diff
@@ -393,14 +393,6 @@ api.compactIri = ({
...
   const inverseCtx = activeCtx.getInverse();
 
-  // if term is a keyword, it can only be compacted to a simple alias
-  if(_isKeyword(iri)) {
-    if(iri in inverseCtx) {
-      return inverseCtx[iri]['@none']['@type']['@none'];
-    }
-    return iri;
-  }
-
```
gkellogg: Another option would be to leave this in place, and exit quickly if an `iri` is found, and otherwise continue. That would allow for more complicated behavior, while keeping the typical case fast.

lib/compact.js
```diff
@@ -393,12 +393,13 @@ api.compactIri = ({
 
   const inverseCtx = activeCtx.getInverse();
 
-  // if term is a keyword, it can only be compacted to a simple alias
-  if(_isKeyword(iri)) {
-    if(iri in inverseCtx) {
-      return inverseCtx[iri]['@none']['@type']['@none'];
-    }
-    return iri;
+  // if term is a keyword, it can may be compacted to a simple alias
```
dlongley: type: `it can may be` => `it may be compacted`

dlongley:
LGTM, just noticed a typo in a comment.

PR #220, opened 2018-01-22
Null map update

gkellogg:
* Index maps on `@none` or an alias
* Improve container ordering for term selection


gkellogg:
Pulling this as builds will fail without.

PR #221, opened 2018-01-25
Compact iri changes

gkellogg:
Only uses terms with IRIs ending in a delim character, or defined with `@prefix` in 1.1 mode when creating compact IRIs.

### DIFF
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -649,7 +649,7 @@ api.compactIri = ({
   for(let i = partialMatches.length - 1; i >= 0; --i) {
     const entry = partialMatches[i];
     const terms = entry.terms;
-    for(let term of terms) {
+    for(const term of terms) {
       // a CURIE is usable if:
       // 1. it has no mapping, OR
       // 2. value is null, which means we're not compacting an @value, AND
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -46,7 +46,8 @@ module.exports = api;
  * @param activeProperty the compacted property associated with the element
  *          to compact, null for none.
  * @param element the element to compact.
- * @param options the compaction options.
+ * @param options the compaction options:
+ *          [issuer] a jsonld.IdentifierIssuer to use to label blank nodes.
  * @param compactionMap the compaction map to use.
  *
  * @return the compacted value.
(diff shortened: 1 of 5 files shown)


lib/compact.js
```diff
@@ -649,14 +649,15 @@ api.compactIri = ({
   for(let i = partialMatches.length - 1; i >= 0; --i) {
     const entry = partialMatches[i];
     const terms = entry.terms;
-    for(let ti = 0; ti < terms.length; ++ti) {
+    for(let term of terms) {
```
dlongley: `const term`

lib/context.js
```diff
@@ -478,6 +484,24 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
...
 
+  // term may be used as a prefix
+  if('@prefix' in value) {
+    if(mapping._termHasColon) {
+      throw new JsonLdError(
+        'Invalid JSON-LD syntax; @context @prefix used on a compact IRI term',
+        'jsonld.SyntaxError',
+        {code: 'invalid term definition', context: localCtx});
+    }
+    if(value['@prefix'] === true || value['@prefix'] === false) {
```
dlongley: Could do `typeof value['@prefix'] === 'boolean'` ... seems more logically clear.

lib/context.js
```diff
@@ -478,6 +484,24 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
...
+      mapping._prefix = value['@prefix'] === true
```
dlongley: Add missing semi-colon.

lib/context.js
```diff
@@ -318,14 +325,13 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
           {code: 'invalid IRI mapping', context: localCtx});
       }
       mapping['@id'] = id;
+      // indicate if this term may be used as a compact IRI prefix
+      mapping._prefix = (!mapping._termHasColon &&
+        id.match(/[:\/\?#\[\]@]$/) &&
```
davidlehn: If this is in a hot path then maybe the regex should be compiled elsewhere.  Or maybe js runtimes are smart enough to do that themselves.

PR #225, opened 2018-01-29
Add `includeRelativeUrls` option to enable relative URLs in RDF output.

dlongley:
This would allow RDF output to include relative URLs. Thoughts?

cc: @davidlehn, @gkellogg, @kidehen

see: https://lists.w3.org/Archives/Public/public-linked-json/2018Jan/0014.html

```js
const jsonld = require('jsonld');

(async () => {

const doc = {
  "@context": {
    "schema": "http://schema.org/",
    "@base": "#"  
  },
  "@id": "#BrewEats",
  "@type": "schema:Restaurant",
  "schema:name": "Brew Eats",
  "databaseId": "23987520"
};

const rdf = await jsonld.toRDF(doc, {
  format: 'application/n-quads',
  includeRelativeUrls: true
});
console.log(rdf);

})();
```

Yielding:

```
<#BrewEats> <http://schema.org/name> "Brew Eats" .
<#BrewEats> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Restaurant> .
```

I'm not sure where the overlap here is with the `produceGeneralizedRdf` flag. I was concerned that simply reusing that flag would do the wrong thing, i.e. people may want relative URLs in the output, but not blank nodes in the predicate position.

### DIFF
--- a/lib/jsonld.js
+++ b/lib/jsonld.js
@@ -638,7 +638,8 @@ jsonld.toRDF = util.callbackify(async function(input, options) {
 
   // set default options
   options = _setDefaults(options, {
-    base: _isString(input) ? input : ''
+    base: _isString(input) ? input : '',
+    includeRelativeUrls: false
   });
 
   // TODO: support toRDF custom map?
--- a/lib/toRdf.js
+++ b/lib/toRdf.js
@@ -62,7 +62,7 @@ api.toRDF = (input, options) => {
         graphTerm = {termType: 'NamedNode'};
       }
       graphTerm.value = graphName;
-    } else {
+    } else if(!options.includeRelativeUrls) {
       // skip relative IRIs (not valid RDF)
       continue;
     }
@@ -105,7 +105,7 @@ function _graphToRDF(dataset, graph, graphTerm, issuer, options) {
         };
 
         // skip relative IRI subjects (not valid RDF)
-        if(!_isAbsoluteIri(id)) {
+        if(!_isAbsoluteIri(id) && !options.includeRelativeUrls) {
           continue;
         }
 
@@ -116,7 +116,7 @@ function _graphToRDF(dataset, graph, graphTerm, issuer, options) {
         };
 
         // skip relative IRI predicates (not valid RDF)
-        if(!_isAbsoluteIri(property)) {
+        if(!_isAbsoluteIri(property) && !options.includeRelativeUrls) {
           continue;
         }
 
@@ -134,7 +134,7 @@ function _graphToRDF(dataset, graph, graphTerm, issuer, options) {
           // convert value or node object to triple
           const object = _objectToRDF(item);
           // skip null objects (they are relative IRIs)
-          if(object) {
+          if(options.includeRelativeUrls || object) {
             dataset.push({
               subject: subject,
               predicate: predicate,


dlongley:
Also, I may need to target something different from master -- in which case we can redo this PR.

gkellogg:
Well, the RDF data model doesn't have relative URLs, although serializations do. However, you can usually coax different parsers and serializers to manage relative URLs. I don't think we can do this in a "standardized" way, but for a jsonld.js feature, that's fine.

kidehen:
@gkellogg, 

I don't understand the comment " RDF data model doesn't have relative URLs" .  RDF simply requires the use of IRIs to identify the subject, predicate, and object (optionally) of a sentence/statement. Thus, a relative or absolute  HTTP URI shouldn't be cause for problems be it input processing or output generation. 

Processors cannot ignore the nature of an HTTP URI i.e., "#" is an integral part of the standard., 

[1] https://www.w3.org/2005/Talks/1110-iswc-tbl/#(7)

gkellogg:
> I don't understand the comment " RDF data model doesn't have relative URLs" . RDF simply requires the use of IRIs to identify the subject, predicate, and object (optionally) of a sentence/statement. Thus, a relative or absolute HTTP URI shouldn't be cause for problems be it input processing or output generation.

From RDF 1.1 Concepts Section 3.2:

> IRIs in the RDF abstract syntax must be absolute, and may contain a fragment identifier.

By “data model”, I was referring to the abstract syntax.

This is described in JSON-LD Section 7 “Relationship to RDF”: https://json-ld.org/spec/latest/json-ld/#relationship-to-rdf.

> Processors cannot ignore the nature of an HTTP URI i.e., "#" is an integral part of the standard.,

“#” is, indeed, a part of the standard as a separating the IRI path from fragment. The IRI “#databaseId” would be a relative IRI evaluated to the current base, when used in a property, that would be the vocabulary base. If no such base is established, it is dropped.

Perhaps we could consider that relative IRIs starting with “#” should be considered to be document-relative, even if they occur in a vocabulary-relative location; that might cause some round-tripping issues. Alternatively, maybe there’s a context syntax to say that the vocabulary base should be taken to be the same as the document base: `"@vocab": "@base"`.

kidehen:
@gkellogg ,

"Perhaps we could consider that relative IRIs starting with “#” should be considered to be document-relative, even if they occur in a vocabulary-relative location; that might cause some round-tripping issues. Alternatively, maybe there’s a context syntax to say that the vocabulary base should be taken to be the same as the document base: "@vocab": "@base"."

That's exactly what a relative HTTP URI is all about. Obscuring this capability is incompatible with the notion of a Semantic Web constructed using Linked Data principles via JSON-LD. 

[1] https://medium.com/openlink-software-blog/simple-linked-data-deployment-tutorial-a532e568c82f -- I should be able to replicate this tutorial using JSON-LD

gkellogg:
Perhaps I'm missing where the tutorial uses relative IRIs for properties. AFAIKT, all examples in the tutorial can be represented as JSON-LD reasonably easily.

kidehen:
@gkellogg, 

Ah! That's correct, I didn't use relative IRIs to denote properties in that post. Thus, we are back to a simpler example re. text doc at: http://kingsley.idehen.net/public_home/kidehen/Public/Linked%20Data%20Documents/Nanotations/basic-test2.txt, where Turtle and JSON-LD tabs should present the same thing but don't currently.

gkellogg:
@kidehen Both playground and my distiller are operating according the the spec. Let's redirect comments on this to json-ld/json-ld.org#488 for how things could change for 1.1.

dlongley:
Closing this, it's not being used and I don't think we want/need the feature.

PR #227, opened 2018-02-08
Framing update

gkellogg:
Update to new JSON-LD 1.1 Framing algorithm.

### DIFF
--- a/lib/expand.js
+++ b/lib/expand.js
@@ -69,6 +69,11 @@ api.expand = ({
     return null;
   }
 
+  // disable framing if activeProperty is @default
+  if(activeProperty === '@default') {
+    options = Object.assign({}, options, {isFrame: false});
+  }
+
   if(!_isArray(element) && !_isObject(element)) {
     // drop free-floating scalars that are not in lists unless custom mapped
     if(!insideList && (activeProperty === null ||
--- a/lib/expand.js
+++ b/lib/expand.js
@@ -8,6 +8,7 @@ const JsonLdError = require('./JsonLdError');
 const {
   isArray: _isArray,
   isObject: _isObject,
+  isEmptyObject: _isEmptyObject,
   isString: _isString
 } = require('./types');
 
@@ -31,7 +32,8 @@ const {
 
 const {
   addValue: _addValue,
-  validateTypeValue: _validateTypeValue
+  validateTypeValue: _validateTypeValue,
+  getValues: _getValues
 } = require('./util');
 
 const api = {};
@@ -206,8 +208,11 @@ api.expand = ({
         'which can be "@type" or "@language".',
         'jsonld.SyntaxError', {code: 'invalid value object', element: rval});
     }
+    const values = rval['@value'] === null ? [] : [].concat(rval['@value']);
+    const types = _getValues(rval, '@type');
+
     // drop null @values unless custom mapped
-    if(rval['@value'] === null) {
+    if(values.length === 0) {
       // TODO: use `await` to support async
       const mapped = expansionMap({
         unmappedValue: rval,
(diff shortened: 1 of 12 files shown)


lib/frame.js
```diff
@@ -202,6 +245,22 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
+      for(const reverseProp of Object.keys(frame['@reverse']).sort()) {
+        const subframe = frame['@reverse'][reverseProp];
+        for(const subject of Object.keys(state.subjects)) {
+          const nodeValues = util.getValues(state.subjects[subject], reverseProp);
+          if(nodeValues.some(v => v['@id'] === id)) {
+            // node has property referencing this subject, recurse
+            output['@reverse'] = output['@reverse'] || {};
+            util.addValue(output['@reverse'], reverseProp, [], {propertyIsArray: true});
+            api.frame(state, [subject], subframe, output['@reverse'][reverseProp], property);
+          }
```
davidlehn: Trailing space (found in another editor).

lib/frame.js
```diff
@@ -443,3 +567,57 @@ function _addFrameOutput(parent, property, output) {
...
+  const t1 = value['@type'];
+  const l1 = value['@language'];
+  const v2 = pattern['@value'] || [];
+  const t2 = pattern['@type'] || [];
+  const l2 = pattern['@language'] || [];
+
+  if(v2.length === 0 && t2.length === 0 && l2.length === 0) {
+    return true;
+  }
+  if(!(v2.includes(v1) || types.isEmptyObject(v2[0]))) {
```
davidlehn: The phantomjs test failed here in some tests where v2 and v1 are the string "P".  Every other test case has v2 as an array.  Based on default above and usage here it looks like that might be a bug?  If not the 'core-js/fn/string/includes' helper would have to be added to main and karma webpack configs.

gkellogg: Yeah, that’s a bug.

lib/frame.js
```diff
@@ -443,3 +567,57 @@ function _addFrameOutput(parent, property, output) {
...
+ * @param pattern used to match value
+ * @param value to check
+ */
+function _valueMatch(pattern, value) {
+  const v1 = value['@value'];
+  const t1 = value['@type'];
+  const l1 = value['@language'];
+  const v2 = pattern['@value'] ? [].concat(pattern['@value']) : [];
+  const t2 = pattern['@type'] ? [].concat(pattern['@type']) : [];
+  const l2 = pattern['@language'] ? [].concat(pattern['@language']) : [];
```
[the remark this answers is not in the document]
gkellogg: Did that, also considered changing it in `expand`, which adds complexity to the expansion algorithm, but reduces even further the needed checks in `_valueMatch`.

lib/expand.js
```diff
@@ -206,8 +213,11 @@ api.expand = ({
         'which can be "@type" or "@language".',
         'jsonld.SyntaxError', {code: 'invalid value object', element: rval});
     }
+    const values = rval['@value'] === null ? [] : [].concat(rval['@value']);
```
davidlehn: Is it worth doing that isArray test here too?

davidlehn:
I'm not sure I know enough to properly review this, but basic code issues look fine.  Something should be done with that new embed api t0005 test.  It should be updated or removed rather than just omitted.  Maybe @dlongley knows more about that.

gkellogg:
I updated the frame-0005-in.jsonld file to be what I think is correct. For some reason, the test runner doesn't actually seem to test against the frame-0005-out.jsonld, and I'm not sure why.

davidlehn:
frame-0005-out.jsonld has that same duplicated ex:hasLibrary property problem that the -in file had.

davidlehn:
Perhaps all those new-embed-api tests should be moved to the main test-suite now?  At least if they don't duplicate other tests.  I think they are all from when the embed API was experimental in jsonld.js.

gkellogg:
I couldn't quite get there through tweaking the output file, so I used the result of the algorithm, which of course, now passes. I think the output is correct, but @dlongley should take a look.

dlongley:
@gkellogg,

I looked over the changed output for that test file ... it seems right to me now unless I'm missing something.

gkellogg:
I think all the nits have been addressed. Any other changes with regard to how framing works, or should work, will impact the spec. I'm going to merge into 0.6.

davidlehn:
playground-dev is now updated with the latest 0.6.x code

PR #228, opened 2018-02-09
Add note about skipping tests.

davidlehn:
- Avoid skipping other potential 1.0 specVersion tests.
- I'm unsure what that test does, but seems it should be skipped explicitly vs skipping all potential 1.0 tests.
- Alternatively to this PR, fix the code so test passes or fix/remove the test.

### DIFF
--- a/tests/test-common.js
+++ b/tests/test-common.js
@@ -29,6 +29,8 @@ const manifest = options.manifest || {
 const TEST_TYPES = {
   'jld:CompactTest': {
     skip: {
+      // skip tests where behavior changed for a 1.1 processor
+      // see JSON-LD 1.0 Errata
       specVersion: ['json-ld-1.0']
     },
     fn: 'compact',


gkellogg:
The `specVersion: 1.0` was added to the test manifest to allow processors that had not updated to continue to run by the 1.0 rules, in this case, for how prefixes are selected for creating compact IRIs. A 1.1 processor, while operating in 1.0 mode will select differently, which is described in the [JSON-LD 1.0 Errata](https://www.w3.org/2001/sw/wiki/JSON_LD_Errata).

When a processor, operating in whatever processing mode, is updated, it will behave differently in such small ways than an unmodified processor. Filtering against `specVersion: 1.0` is intended to provide processors with a way of avoiding obsolete tests.

davidlehn:
Ok, I understand.  I'll change this patch to a comment so future me can figure out why tests are being skipped.  If other specVersion 1.0 tests are added that pass we can revisit just skipping specific tests.

PR #230, opened 2018-02-10
0.6.x work-in-progress

davidlehn:
Tracking PR for 0.6.x work and issues.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -98,7 +98,7 @@ api.process = ({activeCtx, localCtx, options}) => {
           'jsonld.UnsupportedVersion',
           {code: 'invalid @version value', context: ctx});
       }
-      if(activeCtx.processingMode && activeCtx.processingMode.indexOf('json-ld-1.1') !== 0) {
+      if(activeCtx.processingMode && activeCtx.processingMode == 'json-ld-1.0') {
         throw new JsonLdError(
           '@version: ' + ctx['@version'] + ' not compatible with ' + activeCtx.processingMode,
           'jsonld.ProcessingModeConflict',
@@ -260,7 +260,7 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
   const validKeys = ['@container', '@id', '@language', '@reverse', '@type'];
 
   // JSON-LD 1.1 support
-  if(activeCtx.processingMode && activeCtx.processingMode.indexOf('json-ld-1.1') === 0) {
+  if(api.processingMode(activeCtx, 1.1)) {
     validKeys.push('@context');
   }
 
@@ -400,7 +400,7 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
     let hasSet = container.includes('@set');
 
     // JSON-LD 1.1 support
-    if(activeCtx.processingMode && activeCtx.processingMode.indexOf('json-ld-1.1') === 0) {
+    if(api.processingMode(activeCtx, 1.1)) {
       // TODO: @id and @type
       validContainers.push('@graph', '@id', '@type');
 
@@ -819,6 +819,22 @@ api.getContextValue = (ctx, key, type) => {
 api.getAllContexts = async (input, options) => {
   return _retrieveContextUrls(input, options);
 };
+
+/**
+ * Processing Mode check.
+ *
+ * @param activeCtx the current active context.
+ * @param version the or numeric version to check.
+ *
+ * @return boolean.
+ */
+ api.processingMode = (activeCtx, version) => {
+   if(version.toString() >= "1.1") {
+     return activeCtx.processingMode && activeCtx.processingMode >= 'json-ld-' + version.toString();
+   } else {
+     return !activeCtx.processingMode || activeCtx.processingMode == 'json-ld-1.0'
+   }
+ }
 
 /**
  * Returns whether or not the given value is a keyword.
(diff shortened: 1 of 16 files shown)


lib/context.js
```diff
@@ -98,7 +98,7 @@ api.process = ({activeCtx, localCtx, options}) => {
           'jsonld.UnsupportedVersion',
           {code: 'invalid @version value', context: ctx});
       }
-      if(activeCtx.processingMode && activeCtx.processingMode.indexOf('json-ld-1.1') !== 0) {
+      if(activeCtx.processingMode && activeCtx.processingMode == 'json-ld-1.0') {
```
davidlehn: Use ===.

lib/context.js
```diff
@@ -313,14 +331,13 @@ api.createTermDefinition = (activeCtx, localCtx, term, defined) => {
           {code: 'invalid IRI mapping', context: localCtx});
       }
       mapping['@id'] = id;
+      // indicate if this term may be used as a compact IRI prefix
+      mapping._prefix = (!mapping._termHasColon &&
+        id.match(/[:\/\?#\[\]@]$/) &&
+        (simpleTerm || !activeCtx.processingMode || activeCtx.processingMode == 'json-ld-1.0'));
```
davidlehn: Use ===.

lib/context.js
```diff
@@ -792,6 +857,22 @@ api.getAllContexts = async (input, options) => {
...
+ * @param activeCtx the current active context.
+ * @param version the string or numeric version to check.
+ *
+ * @return boolean.
+ */
+ api.processingMode = (activeCtx, version) => {
+   if(version.toString() >= "1.1") {
+     return activeCtx.processingMode && activeCtx.processingMode >= 'json-ld-' + version.toString();
+   } else {
+     return !activeCtx.processingMode || activeCtx.processingMode == 'json-ld-1.0'
```
davidlehn: Use ===.

lib/frame.js
```diff
@@ -90,44 +104,79 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
+    if(id in state.graphMap) {
+      let recurse = false;
+      let subframe = null;
+      if(!('@graph' in frame)) {
+        recurse = state.graph !== '@merged';
+        subframe = {};
+      } else {
+        subframe = frame['@graph'][0];
+        if(!types.isObject(subframe)) {
+          subframe == {};
```
davidlehn: This is clearly not correct.  Should a test cover this somehow?

lib/frame.js
```diff
@@ -315,30 +375,31 @@ function _filterSubjects(state, subjects, frame, flags) {
...
+function _filterSubject(state, subject, frame, flags) {
   // check ducktype
   let wildcard = true;
   let matchesSome = false;
-  for(let key in frame) {
+
+  for(const key in frame) {
+    let matchThis = false;
+    const nodeValues = util.getValues(subject, key);
+    const isEmpty = util.getValues(frame, key).length == 0;
```
davidlehn: Use ===.

lib/frame.js
```diff
@@ -347,34 +408,98 @@ function _filterSubject(subject, frame, flags) {
...
+            return false;
+          }
+          matchThis = true;
+        } else if(frame['@type'].length === 1 && types.isEmptyObject(frame['@type'][0])) {
+          // match on wildcard @type
+          matchThis = nodeValues.length > 0;
+        } else {
+          // match on a specific @type
+          for(const type of frame['@type']) {
+            if(nodeValues.some(tt => tt == type)) {
```
davidlehn: Use ===.

lib/frame.js
```diff
@@ -90,44 +104,79 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
+        }
+        recurse = !(id === '@merged' || id === '@default');
+      }
 
-    // iterate over subject properties
-    let props = Object.keys(subject).sort();
-    for(let i = 0; i < props.length; i++) {
-      const prop = props[i];
+      if(recurse) {
+        state.graphStack << state.graph;
```
davidlehn: I'm guessing this is Ruby code vs a .push()?  Apparently no tests cover this case?

lib/context.js
davidlehn: Or just use `processingMode()`?

lib/context.js
davidlehn: Could `processingMode()` be used here?  Logic is a bit different.

lib/frame.js
gkellogg: It's CYA code, as it should never be the case that `@graph` doesn't have an array of object values. But, perhaps it simply can't happen because of sanity checks in expansion.

lib/context.js
```diff
@@ -792,6 +857,22 @@ api.getAllContexts = async (input, options) => {
...
+/**
+ * Processing Mode check.
+ *
+ * @param activeCtx the current active context.
+ * @param version the string or numeric version to check.
+ *
+ * @return boolean.
+ */
+ api.processingMode = (activeCtx, version) => {
+   if(version.toString() >= "1.1") {
```
davidlehn: Use single quotes.

davidlehn:
@gkellogg Re changelog version header, I've just been leaving that off until right before doing a release.  Since it may change from patch to minor depending on what goes in the release.  And committing to a month (or year!) is just wishful thinking. ;-)

davidlehn:
@gkellogg I update coverage tool in master if you want to pull the change in.  `npm run coverage` works again in node.  karma coverage was causing trouble so I punted on that for now.

dlongley:
@davidlehn -- are we ready to merge this and make a release?

PR #235, opened 2018-03-02
Don't always use arrays for `@graph`.

davidlehn:
Fixes 1.0 compatibility issue.

### DIFF
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -385,9 +385,9 @@ api.compact = ({
               nestResult, itemActiveProperty, compactedItem,
               {propertyIsArray: (!options.compactArrays || container.includes('@set'))});
           } else {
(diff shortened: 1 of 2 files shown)


lib/compact.js
```diff
@@ -385,9 +385,9 @@ api.compact = ({
               nestResult, itemActiveProperty, compactedItem,
               {propertyIsArray: (!options.compactArrays || container.includes('@set'))});
           } else {
-            // wrap using @graph alias, value is always an array
-            if(!_isArray(compactedItem)) {
-              compactedItem = [compactedItem]
+            // wrap using @graph alias, remove array if only one item
+            if(_isArray(compactedItem) && compactedItem.length === 1) {
```
dlongley: Needs `&& !options.compactArrays` too, right?

PR #246, opened 2018-05-22
Add missing await to createNodeMap() and merge().

dmitrizagidulin:
From irc convo with @davidlehn.

### DIFF
--- a/lib/jsonld.js
+++ b/lib/jsonld.js
@@ -705,7 +705,7 @@ jsonld.createNodeMap = util.callbackify(async function(input, options) {
   });
 
   // expand input
-  const expanded = jsonld.expand(input, options);
+  const expanded = await jsonld.expand(input, options);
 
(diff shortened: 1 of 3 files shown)


lib/jsonld.js
```diff
@@ -812,7 +812,7 @@ jsonld.merge = util.callbackify(async function(docs, ctx, options) {
   // compact result (force @graph option to true, skip expansion)
   options.graph = true;
   options.skipExpansion = true;
-  const compacted = jsonld.compact(flattened, ctx, options);
+  const compacted = await jsonld.compact(flattened, ctx, options);
```
dmitrizagidulin: My concern here is - why didn't the `merge()` unit test in `misc.js` catch this line?

davidlehn: The tests don't pass in a context, so the flattened value is returned a few lines up and this code is never run.

PR #257, opened 2018-08-16
Support new WG test suites.

davidlehn:
- Support for json-ld-api tests.
- Future support for json-ld-framing tests.
- Use old json-ld.org framing tests.
- Update document loader to handle w3c URLs.

Note that there are tons of failing tests.  A few are test framework issues that should be addressed before this is merged.


davidlehn:
Split out the valid n-quads patch into a new PR and merged the rest.

PR #258, opened 2018-08-16
Improve active context cache performance.

dlongley:
- Use `Map` to improve cache performance.
- Cache initial contexts based on options.

### DIFF
--- a/lib/ActiveContextCache.js
+++ b/lib/ActiveContextCache.js
@@ -18,9 +18,9 @@ module.exports = class ActiveContextCache {
   }
 
   get(activeCtx, localCtx) {
-    const key = JSON.stringify(localCtx);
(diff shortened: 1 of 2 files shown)


dlongley:
Initial contexts are cached based on `base` and `processingMode` from options. There's no cap on the cache ... so a long running process with many different `base` options would create a large cache. Do we need to address that?

dlongley:
Added a max size for the initial context cache and a dumb implementation to clear it. Can be improved in the future with LRU cache or something if necessary. A bigger improvement at this point would be to avoid doing `JSON.stringify` on every localCtx.

davidlehn:
Merged changes via https://github.com/digitalbazaar/jsonld.js/pull/259

PR #259, opened 2018-08-16
Improve find context urls

dlongley:
Use maps and reduce number of lookups to improve find context URL code. Builds on top of #258.


davidlehn:
Are these changes applicable to pyld too?  May want to add an issue there noting the possible improvement.

davidlehn:
pyld issue added

PR #279, opened 2018-12-12
Update jsonld.js

piggyman007:
I am working on node for ios app (https://code.janeasystems.com/nodejs-mobile) which run on node-chakracore instead of v8.

I found that the spread operator of _setDefaults does not work on node-chakracore, so I would like to change to code by using the basic style (does not affect functional).

Please help to review and approve

### DIFF
--- a/lib/jsonld.js
+++ b/lib/jsonld.js
@@ -1045,10 +1045,15 @@ if(_nodejs) {
   jsonld.useDocumentLoader('xhr');
 }
 
-function _setDefaults(options, {
-  documentLoader = jsonld.documentLoader,
-  ...defaults
-}) {
+function _setDefaults(options, additionalOptions) {
+  const documentLoader = additionalOptions.documentLoader || jsonld.documentLoader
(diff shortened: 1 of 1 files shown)


davidlehn:
I'm not sure what to do here.  We like to be free to use fancy new features.  So far we've basically felt free to use anything in an "Active LTS" Node.js release that has babel support.  Object spread support has been in v8 based Node.js since around 8.3.0.  It's also a stage 4 proposal so should be a standard at some point.  It's entirely possible we'll want to increase object spread usage which would be a problem for ChakraCore.

It appears ChakraCore is the only JS runtime without object spread support?  Seems the future of ChakraCore is uncertain too.  How long would we have to keep such a patch around before updating to the modern syntax?

Are other options like using a babel pre-processor too painful in your case?

piggyman007:
@davidlehn , my code version support both v8 and chakracore.

Is it good idea to merge it to the main branch?

I don't find any problem to ignore it.

PR #287, opened 2019-02-02
Switch to eslint.

davidlehn:
- Current eslint rules unhappy about some testing code.
- Can likely get under control by ignoring files, adjusting rules, and/or adding pragmas.


tests/test-common.js
```diff
@@ -276,7 +276,7 @@ if(options.earl && options.earl.filename) {
   options.earl.report = new EarlReport({id: options.earl.id});
 }
 
-return new Promise((resolve, reject) => {
+return new Promise(resolve => {
```
BigBlueHat: Personally, I like to keep the parenthesis around as it just reads more clearly and when you add a second param you have to put them back anyhow. But that's probably a primarily personal preference. 😃 

The config is [arrow-parens](https://eslint.org/docs/rules/arrow-parens#always) fwiw:
```
/*eslint arrow-parens: ["error", "always"]*/
```

Otherwise, 👍!

.eslintrc.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
BigBlueHat: Curious (only curious 😉), why the choice of `.js` vs. pure JSON?

tests/test-common.js
dlongley: Yeah, there's a permathread in the community on whether or not to always have the parenthesis. So it's a stalemate on this one.

The main counter to "it's easier to refactor" is "basing your style on something *you may do* isn't as good an idea as basing it on what you *do do*". For example, there's a lot of simple functional programming stuff like mapping arrays `foo.map(x => y)` where needlessly adding parenthesis actually makes things harder to read and more annoying to type.

Anyway, we'd sided with that faction ... for now. :smile:

.eslintrc.js
dlongley: Ha! This same debate was just had in the opposite direction ... some argument about keeping everything in the same language (JS) was made. I'm going to stay out of this one.

davidlehn: I do prefer something that allows comments which means yaml or js style.  But otherwise, I don't think it matters in simple declarative use cases like this.

.eslintrc.js
```diff
@@ -0,0 +1,18 @@
...
+    "browser": true,
+    "commonjs": true,
+    "es6": true,
+    "node": true
+  },
+  "extends": "eslint-config-digitalbazaar",
+  "parserOptions": {
+    "ecmaVersion": 2018
+  },
+  "rules": {
```
mattcollier: Are these fixed in the latest release of `eslint-config-digitalbazaar`?

davidlehn: Updated.

tests/graph-container.js
```diff
@@ -74,16 +74,16 @@ describe('@graph container', () => {
       jsonld.compact(expanded, doc['@context'], (err, compacted) => {
         assert.ifError(err);
         assert.deepEqual(compacted, {
-          "@context": {
-            "@version": 1.1,
-            "input": {
-              "@id": "foo:input",
-              "@container": "@graph"
+          '@context': {
```
BigBlueHat: Switching to single quotes means this is no longer copy/paste friendly. 😢 Could you switch object keys back to double quotes?

davidlehn: Yeah, I always debate the wisdom of making otherwise pure json data use any format but json.  I can change this back.

PR #289, opened 2019-02-19
Sealed contexts (plus restrict context nullification to term definitions)

cwebber:
Feature works, and also adds restriction of context nullification to term definitions.  I'm guessing @dlongley will be explaining why this is needed for the extension story to make sense on a call or something soon so I won't bother here.

(Note that I haven't bothered to do tests yet.  I will soon but for the moment I'll just paste some test examples into a comment below.)


cwebber:
Oh yeah and this also BREAKS a test, because, well, it should (since we're adding a new restriction on what's permitted for context nullification).

lib/context.js
```diff
@@ -61,8 +61,15 @@ api.process = ({activeCtx, localCtx, options}) => {
 
     // reset to initial context
     if(ctx === null) {
-      rval = activeCtx = api.getInitialContext(options);
-      continue;
+      if(termContext) {
+        rval = activeCtx = api.getInitialContext(options);
+        continue;
+      } else {
+        throw new JsonLdError(
```
dlongley: Hmm, I don't think this is quite right. We want to allow context nullification anywhere -- the exception is when sealed terms are defined. Sealed terms are not cleared unless via a term definition for a term that is used to extend the JSON tree.

cwebber: That seems surprising to me in that it's introducing a lot more edge-case'y behavior...

dlongley: It shouldn't be -- we're targeting how we clear `@sealed` terms only. `@context` nullification for regular terms is a JSON-LD 1.0 feature that we can't change (for backwards compatibility reasons).

davidlehn:
Add `@sealed` to context.js isKeyword().  (I would have thought that was needed to pass other tests, but I guess not?)

cwebber:
@davidlehn okay I added that one

cwebber:
I'm trying to figure out what @davidlehn wants me to do about these type context nullification rules but my brain is struggling to catch up.  In the meanwhile, how about I paste the manual tests I've been using of what should and shouldn't be working/allowed:

``` javascript
var jsonld = require('./lib/jsonld.js');
var contextLib = require('./lib/context.js');

function printResults(err, expanded) {
    if (err) {
        console.log('error:\n'); console.log(err);
    } else {
        console.log(JSON.stringify(expanded, null, 2));
    }}


// ok
jsonld.expand(
    {'@context': {'foo': {'@id': 'https://example.org/#foo',
                          '@sealed': true},
                  '@version': 1.1,
                  'bar': 'https://example.org/#bar'},
     'foo': 'bar',
     'bar': {'foo': 'baz'}},
    printResults)

// error
jsonld.expand(
    {'@context': {'foo': {'@id': 'https://example.org/#foo',
                          '@sealed': true},
                  '@version': 1.1,
                  'bar': 'https://example.org/#bar'},
     'foo': 'bar',
     'bar': {'@context': {'foo': 'https://schema.org/#oof'},
             'foo': 'baz'}},
    printResults)

// error
jsonld.expand(
    {'@context': {'foo': {'@id': 'https://example.org/#foo'},
                  '@version': 1.1,
                  '@sealed': true,
                  'bar': 'https://example.org/#bar'},
     'foo': 'bar',
     'bar': {'@context': {'foo': 'https://schema.org/#oof'},
             'foo': 'baz'}},
    printResults)

// ok
jsonld.expand(
    {'@context': {'foo': {'@id': 'https://example.org/#foo',
                          '@sealed': false},
                  '@version': 1.1,
                  '@sealed': true,
                  'bar': 'https://example.org/#bar'},
     'foo': 'bar',
     'bar': {'@context': {'foo': 'https://schema.org/#oof'},
             'foo': 'baz'}},
    printResults)

// ok
jsonld.expand(
    {'@context': {'foo': {'@id': 'https://example.org/#foo'},
                  '@version': 1.1,
                  '@sealed': true,
                  'bar': {'@id': 'https://example.org/#bar',
                          '@context': [null]}},
     'foo': 'bar',
     'bar': {'@context': {'foo': 'https://schema.org/#oof'},
             'foo': 'baz'}},
    printResults)
```

cwebber:
Two more things that are not ok

``` javascript
// not ok
jsonld.expand(
    {'@context': {'foo': {'@id': 'https://example.org/#foo'},
                  '@version': 1.1,
                  '@sealed': true,
                  'bar': 'https://example.org/#bar',
                  'Baz': {'@id': 'https://example.org/#Baz',
                          '@context': [null]},
     'foo': 'bar',
     'bar': {'@type': 'Baz',
             '@context': {'foo': 'https://schema.org/#oof'},
             'foo': 'baz'}},
    printResults)

// not ok
jsonld.expand(
    {'@context': {'foo': {'@id': 'https://example.org/#foo'},
                  '@version': 1.1,
                  '@sealed': true,
                  'bar': 'https://example.org/#bar',
                  'Baz': {'@id': 'https://example.org/#Baz',
                          '@context': [null,
                                       {'foo': 'https://example.org/#oof'}]}},
     '@type': 'Baz',
     'foo': 'baz'},
    printResults)
```

lib/context.js
cwebber: Based on our IRC conversation I think I've corrected this to your satisfaction.

davidlehn:
@cwebber: Feel free to add local mocha js tests.  Can add more test files in tests/test.js, and look at misc.js or others as examples.  Or just add to misc.js.  Those are independent files too so easy to run. 
 Sometimes easier to test new features that way before tests are moved to the main test suite.

dlongley:
Yay! Looks like this is doing the right thing now and it was even easier than I thought. I suspect we'll need to add the "propertyTermDefinition" flag to the active context to record its "origin" so that the cache will work properly -- and add some tests that toggle back and forth using the same context to see that it works next.

Looking great, almost entirely done with this feature. I'll need to talk with @gkellogg about it soon.

dlongley:
Do we need to set `propertyTermDefinition: true` here:

https://github.com/digitalbazaar/jsonld.js/blob/1de987cba7470025a409f36162cf03cb7ce10cfa/lib/compact.js#L102-L106

And add some compaction tests for it?

Issue #262, opened 2018-08-25
Use an _asArray() helper.


davidlehn:
Add an `_asArray(data)` helper when code must ensure something is an array.  Use the ternary `Array.isArray(data) ? data : data` construct.  Switch all `[].concat(data)` uses to this.

Simplistic micro benchmark: https://jsperf.com/ensure-array-ternary-vs-concat

Similar to:
https://github.com/ruby-rdf/json-ld/commit/0bd491eccda26bcfc6c01656bbd5a1e084f1f3a0

PR #291, opened 2019-02-20
Add asArray helper.

davidlehn:
Fixes: https://github.com/digitalbazaar/jsonld.js/issues/262

Optimization using an `asArray()` helper.  This doesn't always use a copy like the concat code, but looks to be ok in these use cases.  Has been sitting around for a while.  There was a small measurable difference in some benchmarks IIRC.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Changed
 - Switch to eslint.
+- Optimize ensuring value is an array.
 
 ## 1.5.1 - 2019-02-01
 
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -32,6 +32,7 @@ const {
 
 const {
   addValue: _addValue,
+  asArray: _asArray,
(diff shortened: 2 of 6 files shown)


mattcollier:
I added a bench for this vs another popular pattern we're using in places: https://github.com/digitalbazaar/loop-bench/blob/master/index16.js#L2-L5

davidlehn:
A test I had been using was from https://github.com/davidlehn/json-ld-api/tree/benchmarks (benchmark code there will soon be moving to json-ld.org).  Can run something like the following and switch between master and as-array branches:
```
JSONLD_TESTS=`pwd`/../json-ld-api/tests/benchmarks/b002-manifest.jsonld JSONLD_BENCHMARK=1 npm t
```
Machine I just tested on was having some unrelated load problems, but that particular compaction test was improving from 24k+ ops/s to 26k+ ops/s.  Would be more interesting to test with real data but looks like it does make a difference.  If anyone is interested can run the benchmarks on any test manifest file.

Issue #294, opened 2019-02-28
TypeError: Cannot read property '0' of undefined


koo5:
when trying to parse
```
<file://bn614X>  <http://www.w3.org/1999/02/22-rdf-syntax-ns#first> <http://www.w3.org/1999/02/22-rdf-syntax-ns#nil>.
```

full testcase: 
```js
var done = 0;
(function wait () {if (!done) setTimeout(wait, 1000);})();

const jsonld = require('jsonld');

const line = "<file://bn614X>  <http://www.w3.org/1999/02/22-rdf-syntax-ns#first> <http://www.w3.org/1999/02/22-rdf-syntax-ns#nil>."
jsonld.fromRDF(line, {format: 'application/n-quads'}, (err, doc) => {
            if (err) {
              console.log("err" + err)
            } else {
              console.log("doc"+doc)
            }
            done = true
});

```

davidlehn:
That's a bug in the list handling code.  https://github.com/digitalbazaar/jsonld.js/pull/295 will fix this and make the expanded output should look like:
```json
[
  {
    "@id": "file://bn614X",
    "http://www.w3.org/1999/02/22-rdf-syntax-ns#first": [
      {
        "@list": []
      }
    ]
  }
]
```
I have no idea if that's correct in this case.  I'll add a test to the test suite and get some feedback.

davidlehn:
Fixed in 1.5.4.  Thanks!

PR #295, opened 2019-03-01
Handle "<subject> <rdf:first> <rdf:nil>" triple.

davidlehn:
Fixes https://github.com/digitalbazaar/jsonld.js/issues/294.


davidlehn:
@mattcollier Yes, I proposed test: https://github.com/w3c/json-ld-api/pull/67.  This seems like an edge case.  I'm unsure what the output should be.

Issue #304, opened 2019-03-15
JavaScript keywords are dropped


clecompt-msft:
Vocabulary keys matching JavaScript keywords are incorrectly dropped during expansion, compaction, etc. Consider the following:
```js
require('jsonld').expand({
  "@context": {
    "valueOf": "http://example.org/valueOf",
    "toString": "http://example.org/toString"
  },
  "valueOf": "first",
  "toString": "second"
}).then(result => console.log(JSON.stringify(result, null, 2)))
```
This returns the completely empty...
```json
[]
```
… instead of the expected...
```json
[
  {
    "http://example.org/toString": [
      {
        "@value": "second"
      }
    ],
    "http://example.org/valueOf": [
      {
        "@value": "first"
      }
    ]
  }
]
```

davidlehn:
Yeah, that's a bug.  Somewhat horrifying that such a bug has slipped through for so long!  Thanks for pointing it out.  We'll release some test cases and a fix soon.

It's an issue of using a regular `Object` and `if(term in obj)` vs `if(obj.hasOwnProperty(term))`.  We had wanted to change code using checks like that to use `Map` which would avoid such issues.  But due to some active context implementation details that is a complex change so was put aside.  Will have to just selectively fix it for now.

davidlehn:
Fix is in 1.6.0.  Thanks.

PR #306, opened 2019-03-22
Fix usage of JS Object property names in data.

davidlehn:
Fixes https://github.com/digitalbazaar/jsonld.js/issues/304.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@
 
 ### Changed
 - [rdf-canonize][] updated:
-  - **BREAKING**: A fix was applied that makes the canoncial output format
+  - **BREAKING**: A fix was applied that makes the canonical output format
     properly match the N-Triples canoncial format. This fixes the format to no
     longer escape tabs in literals. This may cause canonical output from
     `jsonld.normalize()`/`jsonld.canonize()` to differ from previous versions
@@ -228,7 +228,7 @@
 ### Changed
 - **BREAKING**: Use RDF JS (rdf.js.org) interfaces for internal
   representation of dataset and quads. This should only break
-  code that was using undocumented internal datastructures,
+  code that was using undocumented internal data structures,
   backwards-compat code exists to handle external RDF parsers.
 - Update `rdf-canonize` to dependency with native support.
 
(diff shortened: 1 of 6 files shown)


lib/context.js
```diff
@@ -686,10 +690,12 @@ api.getInitialContext = options => {
     // create term selections for each mapping in the context, ordered by
     // shortest and then lexicographically least
     const mappings = activeCtx.mappings;
-    const terms = Object.keys(mappings).sort(_compareShortestLeast);
-    for(let i = 0; i < terms.length; ++i) {
-      const term = terms[i];
-      const mapping = mappings[term];
+    const terms = [...mappings.keys()].sort(_compareShortestLeast);
+    for(const term of terms) {
```
mattcollier: would it be more performance to just do `get` first and then check for `undefined` or `null` or whatever?

CHANGELOG.md
```diff
@@ -1,7 +1,21 @@
...
+- Usage of JavaScript Object property names in data.
+  - **NOTE**: A class of bugs was causing term names such as `toString`,
+    `valueOf`, and others to be dropped or produce bogus output. The fix could
+    cause output triples to differ from previous versions if those special
+    names were used.
+  - Specifically, the problem was using `x in obj` instead of
+    `obj.hasOwnProperty(x)` or a `Map`.
+  - Fixed usage in contexts for expand and compact.
+  - Attempted fixes in other parts of the code with similar `x in obj` usage.
+    Finding actual realistic failing test cases proved difficult.
```
mattcollier: I'm not sure this belongs in the changelog, maybe create an issue if positive/negative tests are needed.

lib/context.js
davidlehn: I haven't tested the difference, but it probably is faster when that pattern is ok.  I didn't consider it enough.  I don't think `undefined` is a value so perhaps that could be used.

Might also be faster to sort `mapping.entries()` and iterate through `[key, mapping]` pairs to just avoid `has` and `get` calls.

dlongley:
@davidlehn -- apologies, this has conflicts now. It will need to update the `@protected` stuff to use a `Map`.

PR #309, opened 2019-04-18
Renable doc loader tests

aljones15:
this simply adds a sinon mock for the calls on schema.org so we can unit test the node documentLoader with out the possibility of schema.org being down or the latency.

NOTE: 3 tests are failing on travis ci, but not any of the re-enabled ones.

- adds test-suites to .gitignore
- issues a console.warn if a test makes a real node http call
- re-enables the documentLoader tests
- stubs out an axios call so we can test the default axios calls with out making real ones.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ node_modules
 npm-debug.log
 tests/webidl/*-new
 v8.log
+test-suites/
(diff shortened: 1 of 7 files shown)


aljones15:
@davidlehn this was rebased with master travis ci still has failing tests.

dlongley:
@davidlehn -- can you see if this should be closed or is still relevant?

davidlehn:
Closing.  5.0.0 updated the node document loader tests and uses a local server.  That method allows for testing of various special cases that a server may present.  If there is a desire to use the method here going forward as well, please reopen.

Issue #267, opened 2018-09-25
Remove use of net/global as Angular 6 no longer supports it


Aaron-Sterling:
Angular 6 has removed support for net/global, justification [here](https://github.com/angular/angular-cli/issues/9827#issuecomment-369578814).

This has affected several libraries, including this one. There's a polyfill workaround for this, which I didn't invent, but I describe [here](https://github.com/Aaron-Sterling/ngx-jsonld-provider#step-2-add-polyfill-for-angular-6) in installation instructions to an Angular JSON-LD tool. The long-term recommendation, however, from Angular, is for libraries to remove use of net/global.

Therefore, please remove use of net/global when you can.

dlongley:
I think this has been resolved, closing.

PR #311, opened 2019-05-08
fix: global is unavailable in browser context

chigix:
Signed-off-by: Richard Lea <chigix@zoho.com>


chigix:
Related to the issue in history: https://github.com/digitalbazaar/jsonld.js/issues/267

toyjack (commented):
Worked with vite and vue3! Thanks a lot!

davidlehn:
5.1.0 changed to use `globalThis`. Native support or the bundler/core-js support should work in all browser contexts.

Issue #174, opened 2017-04-23
Tests for `jsonld.promises.compact` do not match the method signature.


mattcollier:
`jsonld.promises.compact` takes two arguments: https://github.com/digitalbazaar/jsonld.js/blob/master/js/jsonld.js#L1274

However, the test suite makes no distinction between the node/promises API with regards to including the 3rd `options` arguments.  Therefore, the test suite calls the `promises.compact` API with three arguments.

This does not lead to a failure, because the method is primisified by using `arguments`: https://github.com/digitalbazaar/jsonld.js/blob/master/js/jsonld.js#L1284

It is unclear how `options` are indented to be passed to the `promises.compact` API.

In fact, if only 2 arguments are given to `promises.compact`, a failure occurs.  This is because the callback added by the `promisify` function ends up being passed to the following function as the `options` parameter: https://github.com/digitalbazaar/jsonld.js/blob/master/js/jsonld.js#L1278 and the `callback` argument is `undefined`.

PR #312, opened 2019-05-08
Fix type scoped terms.

dlongley:
This PR will cause `@type` scoped terms to be limited to objects that have a matching `@type` value. Without this patch, `@type` scoped terms also unexpectedly apply to anything nested (contained within the tree) of a typed object.

### DIFF
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -143,6 +143,9 @@ api.compact = ({
 
     const rval = {};
 
+    // revert type scoped terms
+    activeCtx = activeCtx.revertTypeScopedTerms();
+
     if(options.link && '@id' in element) {
       // store linked element
       if(!options.link.hasOwnProperty(element['@id'])) {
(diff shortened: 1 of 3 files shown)


dlongley:
Addresses #174.

dlongley:
Addresses https://github.com/w3c/json-ld-syntax/issues/174.

gkellogg:
How do you handle expansion test [tm008](https://github.com/w3c/json-ld-api/blob/master/tests/expand/m008-in.jsonld)? Previously, the context passed to expanding values includes those for the type index, but now the expansion algorithm removes those terms.

It seems to me that the object value needs to be changed _before_ it's expanded to include the type index as a value of `@type` (or its alias) so that it will be considered there. This makes type map processing different than for other indexes.

gkellogg:
FWIW, my solution was to pass a `fromMap` flag to expansion when expanding values of a map, which will _note_ clear these term definitions.

dlongley:
@gkellogg,

> FWIW, my solution was to pass a fromMap flag to expansion when expanding values of a map, which will not clear these term definitions.

Sorry, only just now seeing your question -- and I agree with your solution. I was previously passing a flag `insideTypeContainer` that would prevent the clearing of the term definitions, but it wasn't general enough for the new test case you added that combined indexes and type containers. I've updated my implementation to use an `insideIndex` flag that has the same meaning as your `fromMap` flag -- and jsonld.js now passes your new test as well. Thanks!

JFYI, I'm planning on merging this PR on Monday to move things forward with the VC `@context`.

gkellogg:
I’ll have an update to the API PR today, probably won’t get to the syntax. Until after WWW.

Issue #319, opened 2019-07-08
Framing fails in IE11 and Edge with "Invalid calling object"


AlexeyMz:
In IE11 framing fails at `setImmediate()` call, probably due to an Internet Explorer-specific bug which requires `setImmediate()` to be called only on `window` context.

There is already a check related to this issue for IE10 but it doesn't work on IE11 because it defines `setImmediate` function:
https://github.com/digitalbazaar/jsonld.js/blob/master/lib/util.js#L37

EDIT: It also fails in Edge for the same reason.

aljones15:
thanks for this one. if the issue just consistent of adding window to setImmediate you could probably just submit a PR for it.

AlexeyMz:
Hi @aljones15, thanks for a quick reply.

I tried to replace `_setImmediate`-related lines here:
https://github.com/digitalbazaar/jsonld.js/blob/master/lib/util.js#L33
with this:
```
let _setImmediate = typeof setImmediate === 'function' && setImmediate;
try {
  _setImmediate(() => {/* nothing */});
} catch (e) {
  // prevent "Invalid calling object" error in IE and Edge
  _setImmediate = fn => setImmediate(fn);
}
```

And this change fixes the issue in Edge. Unfortunately now it fails in IE with this error:
```
Unhandled promise rejection
SyntaxError: Expected ')'
   at ./node_modules/rdf-canonize/lib/index.js (eval code:2135:1)
   at __webpack_require__ (eval code:25:12)
   at Anonymous function (eval code:46:1)
   at eval code (eval code:1:29)
   at ./lib/jsonld.js (eval code:294:1)
   at __webpack_require__ (eval code:25:12)
   at eval code (eval code:9:1)
   at Anonymous function (eval code:2180:1)
   at __webpack_require__ (eval code:25:12)
   at Anonymous function (eval code:89:11)
```

I think the syntax error is caused by ES6+ syntax in `rdf-canonize`, e.g. async function here: https://github.com/digitalbazaar/rdf-canonize/blob/d410a2f6212073b59b8fe51acdb47de53514e66c/lib/index.js#L82

Should `rdf-canonize` be published in trinspiled form in the NPM package?

aljones15:
Have you tried a polyfil?

https://github.com/YuzuJS/setImmediate
That's one of several.

I'm a bit surprised we don't use one. Thanks for the update.

AlexeyMz:
[...]

PR #321, opened 2019-07-16
Fix "Invalid calling object" in Edge and ES6 syntax error in IE

AlexeyMz:
Included changes:
* Wrap indirect `setImmediate` into function to prevent "Invalid calling object" error in IE/Edge (due to a bug it's an error to call `setImmediate` indirectly).
* Fix `babel-loader` include condition to transpile `rdf-canonize` dependency.

This PR fixes issue #319


AlexeyMz:
Hi @davidlehn @dlongley 
I rebased PR on latest develop to pass the tests. Is there anything else preventing the PR to be merged?

andreikun:
We encountered same issue and we end up importing from dist instead of primary entrypoint something like this: const jsonld = require('jsonld/dist/jsonld.js');
Unfortunately, we still experience "Unhandled promise rejection TypeError: Invalid calling object" and this fix would really help us as we do not want to start using babel loader for 3rd party libraries and node modules.

PR #325, opened 2019-08-28
Support literal JSON.

davidlehn:
work-in-progress.

### DIFF
--- a/lib/expand.js
+++ b/lib/expand.js
@@ -751,7 +751,7 @@ function _expandObject({
   if('@value' in element) {
     const value = element['@value'];
     if(element['@type'] === '@json' && _processingMode(activeCtx, 1.1)) {
-      // Sllow any value, to be verified when the object is fully expanded and the @type is @json.
+      // allow any value, to be verified when the object is fully expanded and the @type is @json.
     } else if((_isObject(value) || _isArray(value)) && !options.isFrame) {
       throw new JsonLdError(
         'Invalid JSON-LD syntax; "@value" value must not be an ' +
--- a/lib/expand.js
+++ b/lib/expand.js
@@ -23,7 +23,8 @@ const {
   expandIri: _expandIri,
   getContextValue: _getContextValue,
   isKeyword: _isKeyword,
-  process: _processContext
+  process: _processContext,
+  processingMode: _processingMode
 } = require('./context');
 
 const {
@@ -289,6 +290,8 @@ api.expand = ({
         'Invalid JSON-LD syntax; only strings may be language-tagged.',
         'jsonld.SyntaxError',
         {code: 'invalid language-tagged value', element: rval});
+    } else if(_processingMode(activeCtx, 1.1) && types.includes('@json') && types.length === 1) {
+      // Any value of @value is okay if @type: @json
     } else if(!types.every(t =>
       (_isAbsoluteIri(t) && !(_isString(t) && t.indexOf('_:') === 0) ||
       _isEmptyObject(t)))) {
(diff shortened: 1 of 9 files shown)


lib/expand.js
```diff
@@ -747,6 +747,20 @@ function _expandObject({
     });
   }
 
+  // @value must not be an object or an array (unless framing) or if @type is @json
+  if('@value' in element) {
+    const value = element['@value'];
+    if(element['@type'] === '@json' && _processingMode(activeCtx, 1.1)) {
+      // Sllow any value, to be verified when the object is fully expanded and the @type is @json.
```
dlongley:
```suggestion
      // allow any value, to be verified when the object is fully expanded and the @type is @json.
```

dlongley: It would be better to combine these conditionals rather than leave this empty one as it's not really a style we use anywhere else.

lib/fromRdf.js
```diff
@@ -285,6 +286,10 @@ function _RDFToObject(o, useNativeTypes) {
     if(!type) {
       type = XSD_STRING;
     }
+    if(type === RDF_JSON_LITERAL) {
+      type = '@json';
+      rval['@value'] = JSON.parse(rval['@value']);
```
dlongley: @davidlehn -- should we do try/catch and wrap in a `JsonLdError` here?

lib/json-canonicalize.js
```diff
@@ -0,0 +1,65 @@
+// https://github.com/cyberphone/json-canonicalization
+// https://www.apache.org/licenses/LICENSE-2.0
+/* eslint-disable */
+
+// ES6 based JSON canonicalizer
+'use strict';
+var canonicalize = function(object) {
```
dlongley: Convert `var` to `const`, etc.

davidlehn: We could update to our style but I left it as-is thinking it would be easier to diff and update later if needed.  Should we partially or fully resstyle it?

davidlehn:
I'm not sure what the guarantees of that json-canonicalize code are.  It depends on the native JSON.stringify.  Do we need some runtime test (maybe from the jcs browser tests) to know if the native code works as expected?  The concern is if this code runs on ye olde browser or odd runtime and the JSON.stringify output differs.

dlongley:
@davidlehn,

> It depends on the native JSON.stringify. Do we need some runtime test (maybe from the jcs browser tests) to know if the native code works as expected?

Sounds like we should add some canonicalization tests over here for that: https://github.com/json-ld/normalization

gkellogg:
> @davidlehn,
> 
> > It depends on the native JSON.stringify. Do we need some runtime test (maybe from the jcs browser tests) to know if the native code works as expected?
> 
> Sounds like we should add some canonicalization tests over here for that: https://github.com/json-ld/normalization

The code is right out of the [RFC](https://tools.ietf.org/id/draft-rundgren-json-canonicalization-scheme-00.html#rfc.appendix.A) example. There's also an [NPM Package](https://www.npmjs.com/package/canonicalize) referenced from the spec to implement this. If you use a package which is tested, I'm not sure what the merits of re-testing are.

davidlehn:
@gkellogg I may be misunderstanding, but I think the algorithm and code depends on an ES6 runtime?  I thought there were some issues if you use older browsers where some behavior was not yet standardized.  Am I mistaken on that point?  I'm just wondering how/if we can detect that so we don't output garbage by mistake if someone runs the code on an old platform.

Also I forgot that code was in the RFC.  Should use that vs the code from the repo.  I also didn't realize that package existed.  NPM search failed me.  Maybe we should just use that.

davidlehn:
From the RFC "JCS relies on serialization of JSON primitives compatible with ECMAScript (aka JavaScript) beginning with version 6".  How do we check, support, or error out on pre-ES6 serializers?

lib/json-canonicalize.js
davidlehn: Fixed by using `canonicalize` package.

lib/fromRdf.js
davidlehn: Wrapped.  Made up names and codes.  Not sure if that is ok or there should be errors for this case in the spec.

davidlehn:
I'm not sure what the best way to detect that the native serializer does what it's supposed to do.  The limited JCS tests appear to run on Node.js 4-12, IE11, and recent Edge, Firefox, Safari, and Chromium.  So we'll call that good enough for now.

PR #326, opened 2019-08-28
Support list of lists.

davidlehn:
Closes: https://github.com/digitalbazaar/jsonld.js/issues/296

### DIFF
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -415,8 +415,6 @@ api.compact = ({
                 relativeTo: {vocab: true}
               })] = expandedItem['@index'];
             }
-          // FIXME
-          //} else if(nestResult.hasOwnProperty(itemActiveProperty)) {
           } else {
             _addValue(nestResult, itemActiveProperty, compactedItem, {
               valueIsArray: true,
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -138,6 +138,22 @@ api.compact = ({
       return rval;
     }
 
+    // if expanded property is @list and we're contained within a list
+    // container, recursively compact this item to an array
+    if(_isList(element)) {
+      const container = _getContextValue(
+        activeCtx, activeProperty, '@container') || [];
+      if(container.includes('@list')) {
+        return api.compact({
+          activeCtx,
+          activeProperty,
+          element: element['@list'],
+          options,
+          compactionMap
+        });
+      }
+    }
+
     // FIXME: avoid misuse of active property as an expanded property?
     const insideReverse = (activeProperty === '@reverse');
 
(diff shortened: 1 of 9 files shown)


lib/nodeMap.js
```diff
@@ -64,7 +64,7 @@ api.createNodeMap = (input, graphs, graph, issuer, name, list) => {
   // add values to list
   if(graphTypes.isValue(input)) {
     if('@type' in input) {
-      let type = input['@type'];
+      const type = input['@type'];
```
davidlehn: A few lines down `type` can change so needs to be a `let` here.  Maybe test coverage doesn't hit that conditional.

lib/nodeMap.js
```diff
@@ -171,9 +172,7 @@ api.createNodeMap = (input, graphs, graph, issuer, name, list) => {
       util.addValue(subject, property, [], {propertyIsArray: true});
       continue;
     }
-    for(let oi = 0; oi < objects.length; ++oi) {
-      let o = objects[oi];
-
+    for(var o of objects) {
```
davidlehn:
```suggestion
    for(let o of objects) {
```

lib/nodeMap.js
```diff
@@ -64,7 +64,7 @@ api.createNodeMap = (input, graphs, graph, issuer, name, list) => {
   // add values to list
   if(graphTypes.isValue(input)) {
     if('@type' in input) {
-      const type = input['@type'];
+      let type = input['@type'];
```
gkellogg: The fact that this did not trigger when running tests implies a missing test case.

davidlehn: Yes. "npm run coverage" is helpful to find such things. It does show that line not being executed.

lib/compact.js
```diff
@@ -399,14 +415,14 @@ api.compact = ({
...
-          } else if(nestResult.hasOwnProperty(itemActiveProperty)) {
-            // can't use @list container for more than 1 list
-            throw new JsonLdError(
-              'JSON-LD compact error; property has a "@list" @container ' +
-              'rule but there is more than a single @list that matches ' +
-              'the compacted term in the document. Compaction might mix ' +
-              'unwanted items into the list.',
-              'jsonld.SyntaxError', {code: 'compaction to list of lists'});
+          // FIXME
+          //} else if(nestResult.hasOwnProperty(itemActiveProperty)) {
```
davidlehn: @gkellogg Can this FIXME and conditional be removed?  I think I was adapting the ruby code changes and wasn't sure what to do here.

lib/fromRdf.js
```diff
@@ -225,22 +225,23 @@ api.fromRDF = async (
...
-
-        // preserve list head
-        if(RDF_REST in graphObject[head['@id']]) {
-          head = graphObject[head['@id']][RDF_REST][0];
-        }
-
-        list.pop();
-        listNodes.pop();
-      }
+      // FIXME
```
davidlehn: @gkellogg I think this is another case of trying to adapt the ruby changes. Is this block needed anymore?

lib/nodeMap.js
```diff
@@ -190,6 +189,13 @@ api.createNodeMap = (input, graphs, graph, issuer, name, list) => {
           subject, property, {'@id': id},
           {propertyIsArray: true, allowDuplicate: false});
         api.createNodeMap(o, graphs, graph, issuer, id);
+      } else if(graphTypes.isValue(o)) {
+        //if('@type' in o) {
+        //  o['@type'] = o['@type'][0];
+        //}
```
davidlehn: I don't recall what this commented code was for. Anyone know if it is needed?

gkellogg: It seemed to be for a case that can't happen, as it's handled later in the algorithm, so this would never happen.

lib/compact.js
gkellogg: My code doesn't have such a case, and it doesn't seem to be necessary for passing tests; I don't know what the original intention was for.

lib/fromRdf.js
gkellogg: My code has the same commented-out bit, and I don't recall the original motivation. I think the test suite is comprehensive here, so the code doesn't seem necessary in either library.

gkellogg: We have adequate tests for list of lists in from Rdf, so this can be dropped.

PR #332, opened 2019-10-27
Remove callback support and make process context async.

dlongley:
This PR removes deprecated callback support with an aim towards supporting `options` in document loaders. It also makes processing contexts asynchronous with an aim towards allowing contexts to load during processing and to enable better active context caching and reuse based on contextual information during context processing.


lib/compact.js
```diff
@@ -51,9 +51,9 @@ module.exports = api;
  * @param options the compaction options.
  * @param compactionMap the compaction map to use.
  *
- * @return the compacted value.
+ * @return a promise that resolves to the compacted value.
  */
-api.compact = ({
+api.compact = async ({
```
mattcollier: I assume  this maps directly to `jsonld.compact` in which case this breaking change from sync to async warrants a mention in the changelog.

lib/expand.js
```diff
@@ -63,7 +63,7 @@ module.exports = api;
  *
  * @return a Promise that resolves to the expanded value.
  */
-api.expand = ({
+api.expand = async ({
```
mattcollier: looks like another important change for the changelog.

lib/compact.js
dlongley: It does not map directly. `jsonld.compact` was already async. The internals were not -- so this is not a public facing change. Same for the other internals.

lib/expand.js
dlongley: See above. These are internal functions that were previously synchronous. The external ones were already asynchronous. That being said, there *is* a separate change that removes callback support on the public API, but the CHANGELOG covers that already.

dlongley:
@mattcollier,

> Does the readme need to be updated before release as well?

Yes, but I don't think we'll do a release immediately after this PR is in. I have a few other changes I'm working on that I want to get into the release that are based off of this PR.

tests/graph-container.js
```diff
@@ -18,8 +18,11 @@ describe('@graph container', () => {
         "value": "x"
       }
     };
-    jsonld.expand(doc, (err, expanded) => {
-      assert.ifError(err);
+    const p = jsonld.expand(doc);
+    assert(p instanceof Promise);
+    p.catch(e => {
```
davidlehn: Could change many/all of these tests to async functions and return a promise.  Avoids needing to call done().

dlongley:
Updated README.

tests/graph-container.js
dlongley: @davidlehn, yup, but was just copying the convention in the file. Another PR could do that clean up later.

lib/jsonld.js
```diff
@@ -613,34 +605,10 @@ jsonld.fromRDF = util.callbackify(async function(dataset, options) {
...
-        }
-      });
-    });
-  } else {
-    parsedDataset = Promise.resolve(rdfParser(dataset));
-  }
-
-  parsedDataset = await parsedDataset;
-
-  // back-compat with old parsers that produced legacy dataset format
```
davidlehn: Is this removed on purpose, and if so, should it have a changelog comment?

dlongley: I added a note about it.

README.md
```diff
@@ -181,120 +181,87 @@ const context = {
...
 
 ```js
 // deserialize N-Quads (RDF) to JSON-LD
-jsonld.fromRDF(nquads, {format: 'application/n-quads'}, (err, doc) => {
-  // doc is JSON-LD
-});
-
-// or using promises
-const doc = await jsonld.fromRDF(nquads, {format: 'application/n-quads'});
+const doc = jsonld.fromRDF(nquads, {format: 'application/n-quads'});
```
davidlehn: await?

dlongley: Added

dlongley:
@davidlehn, ok, I'm merging this. If you think we need more information on how to upgrade, please add in another PR.

PR #334, opened 2019-11-15
Updates to indexing

gkellogg:
* Indexing on `@type` requires `@type` to be either `@id` or `@vocab`,  and defaults to `@id`.
* Indexing on an arbitrary property, not just `@index`.

### DIFF
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -521,7 +521,31 @@ api.compact = async ({
             }
             key = expandedItem['@language'];
           } else if(container.includes('@index')) {
-            key = expandedItem['@index'];
+            const indexKey = _getContextValue(activeCtx, itemActiveProperty, '@index') || '@index';
+            const containerKey = api.compactIri({activeCtx, iri: indexKey, vocab: true});
+            if(indexKey === '@index') {
+              key = expandedItem['@index'];              
+              delete compactedItem[containerKey];
+            } else {
+              let others;
+              [key, ...others] = _asArray(compactedItem[indexKey] || []);
+              if(!_isString(key)) {
+                // Will use @none if it isn't a string.
+                key = null;
+              } else {
+                switch(others.length) {
+                  case 0:
+                    delete compactedItem[indexKey];
+                    break;
+                  case 1:
+                    compactedItem[indexKey] = others[0];
+                    break;
+                  default:
+                    compactedItem[indexKey] = others;
+                    break;
+                }
+              }
+            }
           } else if(container.includes('@id')) {
             const idKey = api.compactIri({activeCtx, iri: '@id', vocab: true});
             key = compactedItem[idKey];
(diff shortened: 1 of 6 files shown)


lib/compact.js
```diff
@@ -544,6 +544,18 @@ api.compact = async ({
                 compactedItem[typeKey] = types;
                 break;
             }
+
+            // If compactedItem contains a single entry
+            // whose key maps to @id, recompact without @type
+            if(Object.keys(compactedItem).length === 1 && '@id' in expandedItem) {
+              compactedItem = await api.compact({
```
dlongley: Can we just get the alias for `@id` here and pull out the value `compactedItem[alias]` without having to run compact again?

dlongley:
Modulo my comment this PR seems ok to me so far. @davidlehn ?

lib/compact.js
gkellogg: Not easily, as if reduced to a node reference, it can be turned into a string. If not, it remains a node object. We could conceivably implement that logic there, but that's what expand is for, and it should be pretty quick in that case.

lib/context.js
```diff
@@ -595,6 +609,21 @@ api.createTermDefinition = (
     mapping['@container'] = container;
   }
 
+  // property indexing
+  if('@index' in value) {
+    if (!('@container' in value) || !mapping['@container'].includes('@index')) {
```
davidlehn: Minor style nit:
```suggestion
    if(!('@container' in value) || !mapping['@container'].includes('@index')) {
```

lib/context.js
```diff
@@ -595,6 +609,21 @@ api.createTermDefinition = (
...
+      throw new JsonLdError(
+        'Invalid JSON-LD syntax; @index without @index in @container: ${value["@index"]} on term ${term}.', 'jsonld.SyntaxError',
+        {code: 'invalid term definition', context: localCtx});
+    }
+    if (!_isString(value['@index']) || value['@index'].indexOf('@') === 0) {
```
davidlehn: Minor style nit:
```suggestion
    if(!_isString(value['@index']) || value['@index'].indexOf('@') === 0) {
```

lib/context.js
```diff
@@ -595,6 +609,21 @@ api.createTermDefinition = (
     mapping['@container'] = container;
   }
 
+  // property indexing
+  if('@index' in value) {
+    if (!('@container' in value) || !mapping['@container'].includes('@index')) {
+      throw new JsonLdError(
+        'Invalid JSON-LD syntax; @index without @index in @container: ${value["@index"]} on term ${term}.', 'jsonld.SyntaxError',
```
davidlehn: line length

davidlehn: same below

lib/expand.js
```diff
@@ -610,14 +610,19 @@ async function _expandObject({
...
+
       expandedValue = await _expandIndexMap({
         activeCtx: termCtx,
         options,
         activeProperty: key,
         value,
         expansionMap,
         asGraph,
-        indexKey: '@index'
+        indexKey: indexKey,
```
davidlehn:
```suggestion
        indexKey,
```

PR #337, opened 2019-11-25
@context resolution respects relative URLs.

ericprud:
c.f. w3c/json-ld-api#228

### DIFF
--- a/lib/documentLoaders/node.js
+++ b/lib/documentLoaders/node.js
@@ -83,7 +83,7 @@ module.exports = ({
     const statusText = http.STATUS_CODES[res.statusCode];
     if(res.statusCode >= 400) {
       throw new JsonLdError(
-        'URL ' + url + ' could not be dereferenced: ' + statusText,
+        `URL "${url}" could not be dereferenced: ${statusText}`,
         'jsonld.InvalidUrl', {
           code: 'loading document failed',
           url,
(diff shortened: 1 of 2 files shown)


gkellogg:
LGTM.

lib/documentLoaders/node.js
```diff
@@ -83,7 +83,7 @@ module.exports = ({
     const statusText = http.STATUS_CODES[res.statusCode];
     if(res.statusCode >= 400) {
       throw new JsonLdError(
-        'URL could not be dereferenced: ' + statusText,
+        'URL ' + url + ' could not be dereferenced: ' + statusText,
```
dlongley:
```suggestion
        `URL "${url}" could not be dereferenced: ${statusText}`,
```

lib/documentLoaders/node.js
```diff
@@ -83,7 +83,7 @@ module.exports = ({
     const statusText = http.STATUS_CODES[res.statusCode];
     if(res.statusCode >= 400) {
       throw new JsonLdError(
-        'URL could not be dereferenced: ' + statusText,
+        `URL "${url}" could not be dereferenced: ${statusText}`,
```
davidlehn: This is a change from our usual style.  URL is already a param in the details so could be omitted in the message text.  Could also move the statusText to details.

davidlehn: I'll merge PR, but may change this text later.

PR #338, opened 2019-12-02
Allow compaction of a single type to an array.

gkellogg:
`@type` may be used as a term definition only if `"@container": "@set"`.

### DIFF
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -24,7 +24,8 @@ const {
   expandIri: _expandIri,
   getContextValue: _getContextValue,
   isKeyword: _isKeyword,
-  process: _processContext
+  process: _processContext,
+  processingMode: _processingMode
 } = require('./context');
 
 const {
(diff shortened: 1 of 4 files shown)


lib/compact.js
```diff
@@ -220,7 +221,15 @@ api.compact = async ({
...
-        const isArray = _isArray(compactedValue) && expandedValue.length === 0;
+        const container = _getContextValue(
+          activeCtx, alias, '@container') || [];
+
+        // treat as array for @type if @container includes @set
+        const typeAsSet = isType &&
+          container.includes('@set') &&
+          _processingMode(activeCtx, 1.1);
+        const isArray =
+          (_isArray(compactedValue) && expandedValue.length === 0) || typeAsSet;
```
davidlehn: Potential micro optimization?  Check `typeAsSet` first since it's already calculated and maybe can short circuit the other tests.  Double check this logic please.
```suggestion
          typeAsSet || (_isArray(compactedValue) && expandedValue.length === 0);
```

Issue #339, opened 2019-12-03
Do not process `@context` in JSON literals


dlongley:
We currently process `@context` anywhere it appears in a document in a pre-processing step. This breaks JSON literals. I'm working on a patch to do JIT context processing instead -- with a number of other caching improvements. This patch should also fix this bug.

davidlehn:
Is a main test suite test needed for variations of this or is that too specific of an implementation detail of what not to do?
cc: @gkellogg

gkellogg:
I think anytime an implementation encounters an issue that might not be covered by the test suite and others might make it, it is appropriate to add such a test to the suite. 

If you’d like to contribute such a test, that would be great.

dlongley:
Addressed by #342.

PR #342, opened 2019-12-07
Implement just-in-time context resolution.

dlongley:
This PR removes the preprocessing step for resolving contexts and replaces its functionality with JIT context resolution. This should fix problems with mutating JSON literal content. This feature also allows document loaders to return an additional `tag` property that jsonld.js can use to more efficiently cache and reuse already processed contexts.

- Add optional `tag` feature processing to returned RemoteDocuments.
  A `tag` will be understood to mean that the same context document
  needn't be processed twice. A special `tag` value of `static` is
  interpreted to mean that a context does not even need to be retrieved
  more than once.
- Enables greater reuse of already processed contexts and quicker
  discovery (via `static` tag) of already processed contexts for
  a given context URL (instead of requiring the context content
  itself to be seached for and found in a cache, only its URL
  needs to be found).
- Addresses #339.


lib/ContextResolver.js
```diff
@@ -0,0 +1,243 @@
...
+    // resolve each context in the array
+    const allResolved = [];
+    for(const ctx of context) {
+      if(_isString(ctx)) {
+        // see if `ctx` has been resolved before...
+        let resolved = this._get(ctx);
+        if(!resolved) {
+          // not resolved yet, resolve
+          resolved = await this._resolveRemoteContext(
+            {url: ctx, documentLoader, base, cycles});
```
gkellogg: Does this have the affect of serializing all remote context fetches?

lib/ContextResolver.js
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    // track cycles
+    cycles.add(url);
+
+    let context;
+    let remoteDoc;
+
+    try {
+      remoteDoc = await documentLoader(url);
```
gkellogg: Note that we're going to need to pass in API arguments to the documentLoader for options such as `extractAllScripts`, but this can be handled in a different PR.

dlongley:
@gkellogg,

> Presumably, base is handled properly per recent tests.

The tests pass -- including the `URL resolution follows RFC3986` ones.

lib/context.js
```diff
@@ -65,24 +60,34 @@ api.process = async ({
...
 
-  // override propagate if localCtx has `@propagate`
-  if(_isObject(ctxs[0]) && typeof ctxs[0]['@propagate'] === 'boolean') {
-    // Retrieve early, error checking done later
-    propagate = ctxs[0]['@propagate'];
+  // resolve contexts
+  const resolved = await options.contextResolver.resolve({
+    context: localCtx,
+    documentLoader: options.documentLoader,
+    base: options.base
```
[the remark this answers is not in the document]
dlongley: I don't think it's ever `@base` (could be wrong). It is either a document location (if the API was invoked with a URL) or the base passed via the API options. The tests pass.

lib/ContextResolver.js
dlongley: `_resolveRemoteContext` will fetch all remote contexts (recursively) and return an array of resolved contexts. Note that this does not include any scoped contexts, i.e., it doesn't deeply inspect the resolved contexts for those, but it will resolve any relative URLs encountered in scoped contexts to ensure the base URL used is proper. Any scoped contexts will be resolved later, JIT.

lib/ContextResolver.js
dlongley: Yeah. I didn't want to do anything with that in this PR but it will need to be addressed in a subsequent PR.

lib/jsonld.js
```diff
@@ -90,6 +91,11 @@ const wrapper = function(jsonld) {
 /** Registered RDF dataset parsers hashed by content-type. */
 const _rdfParsers = {};
 
+// resolved context cache
+// TODO: consider basing max on context size rather than number
+const RESOLVED_CONTEXT_CACHE_MAX_SIZE = 100;
```
davidlehn: Is this adjustable?  Should there be some API so it can be adjustable or a custom cache could be used?

lib/ContextResolver.js
```diff
@@ -0,0 +1,243 @@
...
+      {url, documentLoader, cycles});
+
+    // update base according to remote document and resolve any relative URLs
+    base = remoteDoc.documentUrl || url;
+    _resolveContextUrls({context, base});
+
+    // resolve, cache, and return context
+    const resolved = await this.resolve(
+      {context, documentLoader, base, cycles});
+    this._cacheResolvedContext({key: url, resolved, tag: remoteDoc.tag});
```
davidlehn: Need some docs on the tag feature.  Is 'static' special?

lib/jsonld.js
dlongley: It was not adjustable in the previous version, so there's no change here. We can make it so in a future PR.

lib/ContextResolver.js
dlongley: `static` is a special value, yes. We can document in a separate PR.

lib/ContextResolver.js
dlongley: If by `this` you meant "the `await`" -- then, yes, an array of context URLs will be loaded serially now. A future PR can add more complexity here to parallelize that.

PR #346, opened 2019-12-10
Fix JSON literal value version 1 check.

dlongley:
So this fixes a bug where the JSON literal version 1 check was being performed on the unexpanded input instead of on the expanded output. IOW, it was looking for `@value` on the unexpanded input and then doing some version 1 check on it if `@type` was not `@json`. If the input uses an alias for `@type` (like `type`) this check will be erroneously triggered and throw an error. Example input:

```
{
  "@context": {
    "type": "@type"
  },
  "ex:foo": {
    "type": "@json",
    "@value": {test: 1}
  }
}
```

The conditional should have been checking for `@type` on the expanded output (`expandedParent`) instead -- a similar problem would have occurred if `@value` had been aliased.

However, in fixing this bug we start to fail some test:

`#te029 Verifies that an exception is raised in Expansion when an invalid value object value is found`

In looking at the input for that test, however, it seems like the test is wrong. The input is:

```
{
  "@type": true
}
```

Which is not a value object (there is no `@value` ... it is a top-level input). So I'm not quite sure what should be done here as I don't know what that test was suppose to test originally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # jsonld ChangeLog
 
+### Fixed
+- JSON literal value handling issues.
+
 ## 2.0.0 - 2019-12-09
 
 ### Notes
--- a/lib/expand.js
+++ b/lib/expand.js
@@ -406,6 +406,7 @@ async function _expandObject({
 }) {
   const keys = Object.keys(element).sort();
   const nests = [];
+  let unexpandedValue;
   for(const key of keys) {
     let value = element[key];
     let expandedValue;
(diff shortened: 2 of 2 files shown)


davidlehn:
@dlongley That input you mention is for e028-in.jsonld, not e029-in.jsonld.

dlongley:
> @dlongley That input you mention is for e028-in.jsonld, not e029-in.jsonld.

Oh, then maybe everything is fine!

davidlehn:
Not fine, test still fails.  The expandedParent change isn't quite right.  The `@value` checks need to be on the unexpanded data.  Can fix by reverting the `expandedParent` change back to `element` except for that one that accesses `@type`.  (Tested that with the new type and value alias tests.)  Is that more correct?

dlongley:
@davidlehn -- it seems like the check would just get skipped if you looked for `@value` on the unexpanded data and `value` alias were used.

PR #347, opened 2019-12-28
Load html

gkellogg:
Load JSON-LD from HTML documents.

* Adds options parameter to documentLoader
* Uses xmldom, if loaded.
* Adds util.ParseContentTypeHeader
* Adds documentLoader implementations for xhr and node (still requires tests).
* Set default for expandAllScripts to true for flatten and toRdf.


gkellogg:
Some tests still required for XHR and Node documentLoaders.

Some work needed to optionally append text/html (and application/xhtml+xml) to accept headers if we can parser HTML.

Perhaps some abstracted parsing support to handle other HTML parsers?

Otherwise, please see if you agree with the general direction of this work.

lib/jsonld.js
```diff
@@ -865,7 +871,10 @@ jsonld.get = async function(url, options) {
     load = jsonld.documentLoader;
   }
 
-  const remoteDoc = await load(url);
+  // FIXME: unescape frag?
```
gkellogg: Not sure if this is necessary; it's not to pass tests.

lib/jsonld.js
```diff
@@ -874,16 +883,68 @@ jsonld.get = async function(url, options) {
...
+          // only application/ld+json
+          if(!script.getAttribute('type').startsWith('application/ld+json')) {
+            continue;
+          }
+          // If url has a fragment identifier, only matching scripts
+          if(frag && script.getAttribute('id') !== frag) {
+            continue;
+          }
+          try {
+            remoteDoc.document.push(JSON.parse(script.textContent));
```
gkellogg: Note that `textContent` always decodes entities, so there are tests we can't pass. Any ideas?

lib/jsonld.js
```diff
@@ -934,6 +995,20 @@ jsonld.documentLoaders = {};
 jsonld.documentLoaders.node = require('./documentLoaders/node');
 jsonld.documentLoaders.xhr = require('./documentLoaders/xhr');
 
+// Optional DOM parser
+try {
+  jsonld.domParser = require('xmldom').DOMParser;
```
gkellogg: This will support xmldom, if it's loaded. Are there other parsers that should be included? How to configure and test?

lib/util.js
```diff
@@ -15,6 +15,7 @@ const REGEX_LINK_HEADER = /\s*<([^>]*?)>\s*(?:;\s*(.*))?/;
 const REGEX_LINK_HEADER_PARAMS =
   /(.*?)=(?:(?:"([^"]*?)")|([^"]*?))\s*(?:(?:;\s*)|$)/g;
 
+// FIXME: conditinally support text/html
```
gkellogg: With HTML support, we should include `text/html` and `application/ld+json`, but not sure best way to do that. It will also mess up some documentLoader tests?

lib/documentLoaders/xhr.js
```diff
@@ -30,7 +34,7 @@ module.exports = ({
   const queue = new RequestQueue();
   return queue.wrapLoader(loader);
 
-  async function loader(url) {
+  async function loader(url, options) {
```
gkellogg: There don't seem to be standalone XHR documentLoader tests, as there are node tests.

gkellogg:
Perhaps the webpack build is failing because it doesn't include xmldom? Should there be a way to condition tests on this? How?

dlongley:
Thanks @gkellogg! This is great. A few minor changes needed plus one architectural thing we need some feedback from @davidlehn on (regarding how to pull in a dom parser).

lib/jsonld.js
BigBlueHat: FWLIW, I threw together a JSBin that uses [example 10](https://w3c.github.io/json-ld-api/tests/html/e010-in.html) to test parsing via jsonld.js 2.0.2:
https://jsbin.com/rewaxiquki/edit?html,console

There aren't any entity decoding issues, so I think this issue is probably a bug in `xmldom`'s implementation.

Here's another one which works using [jsdom](http://npmjs.com/package/jsdom):
https://runkit.com/embed/9qrv8dl5g2bs

lib/util.js
```diff
@@ -142,6 +143,36 @@ api.parseLinkHeader = header => {
...
+ * Accept: application/ld+json;profile=http://www.w3.org/ns/json-ld#context
+ *
+ * Parses as: ["application/ld+json",
+ *             {profile: "http://www.w3.org/ns/json-ld#context"}]
+ *
+ * If there is more than one
+ *
+ * @param header the content-type header to parse.
+ */
+api.parseContentTypeHeader = header => {
```
davidlehn: - Docs above are outdated, mentioning Accept and rel.
- Perhaps this should just use https://github.com/jshttp/content-type. It's a bit bigger but far more battle tested than a custom solution.

gkellogg: Okay, updated to use that package.

gkellogg:
@davidlehn are we waiting for something else to merge this?

lib/documentLoaders/xhr.js
```diff
@@ -3,7 +3,12 @@
  */
 'use strict';
 
-const {parseLinkHeader, buildHeaders} = require('../util');
+const contentType = require ('content-type');
```
davidlehn:
```suggestion
const contentType = require('content-type');
```

lib/jsonld.js
```diff
@@ -34,6 +34,7 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 const canonize = require('rdf-canonize');
+const contentType = require ('content-type');
```
davidlehn:
```suggestion
const contentType = require('content-type');
```

lib/documentLoaders/node.js
```diff
@@ -3,7 +3,12 @@
  */
 'use strict';
 
-const {parseLinkHeader, buildHeaders} = require('../util');
+const contentType = require ('content-type');
```
davidlehn:
```suggestion
const contentType = require('content-type');
```

davidlehn:
@gkellogg I was waiting on me to look into how to improve the way the `xmldom` lib is used.  Otherwise I think patch is good.  As Dave's comment mentioned, might want some sort of registered parser pattern.  Need to check how the current code works with bundlers and such.

lib/jsonld.js
rubensworks: FYI, I use [htmlparser2](https://www.npmjs.com/package/htmlparser2) in my implementation, and it doesn't seem to be decoding entities, so it allows all tests to pass.
So I also think that this is a bug/feature in xmldom.

lib/jsonld.js
```diff
@@ -934,6 +995,27 @@ jsonld.documentLoaders = {};
 jsonld.documentLoaders.node = require('./documentLoaders/node');
 jsonld.documentLoaders.xhr = require('./documentLoaders/xhr');
 
+// Optional DOM parser
+try {
+  jsonld.domParser = require('xmldom').DOMParser || class NoDOMParser {
```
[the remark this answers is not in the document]
dmitrizagidulin: Update: PR https://github.com/digitalbazaar/jsonld.js/pull/341 removes `xmldom` as a dependency.

This PR here (#347) should be updated to take that into account (specifically line 1022 here).

PR #348, opened 2019-12-30
JSON literals which are null.

gkellogg:


lib/expand.js
```diff
@@ -402,11 +407,20 @@ async function _expandObject({
...
   const keys = Object.keys(element).sort();
   const nests = [];
   let unexpandedValue;
+
+  // Remember inputType to aid with JSON literals
+  let inputType = element[typeKey];
+  if(inputType) {
+    inputType = _asArray(inputType).slice(-1)[0];
+    inputType = _expandIri(activeCtx, inputType, {vocab: true}, options);
+  }
```
davidlehn: Funny code smell here.  In the end it's just testing if `inputType` is `@json`.  Does that only happen when `element[typeKey]` is either `"@json"` or `["@json"]` (or a `@json` alias)?  I think other inputs would be another error?  A bit of a thinko to rework the logic, but seems like `_expandIri` should only run in those two cases, else set `inputType` to `null` or similar.

gkellogg: Its true that the only use of `inputType` is to see if it is `@json`, but _expandIri needs to be run on something, and as type could be an array or a string, something's needed to see if it resolves to `@type`: `@json`. I'm open to suggestions.

This is a bit easier in Ruby, at least for me.

gkellogg: Is this more acceptable?

```js
  if(element[typeKey]) {
    var inputType =
      _expandIri(activeCtx,
        _asArray(element[typeKey]).slice(-1)[0],
        {vocab: true}, options);
  }
```

gkellogg: Might be better using a ternary operator:

```js
const inputType = element['typeKey'] ?
  _expandIri(activeCtx,
        _asArray(element[typeKey]).slice(-1)[0],
        {vocab: true}, options) :
  null
```

davidlehn: I was suggesting code should avoid the `_expandIri` call unless you have a string or a single element array.  But maybe that's pointless since `_expandIri` also does checks that would end up duplicated without some code refactoring.  When `@json` is being used, I couldn't figure out a valid input that was anything but `"@json"` or `["@json"]` (or with an alias).  If those are in fact the only two cases that need to be checked, the `_asArray` could be avoided with a quick check.  Also the `@json` check could be here and the flag used later:
```js
  let inputType = element[typeKey];
  if(inputType) {
    if(_isArray(inputType) && inputType.length === 1) {
      inputType = inputType[0];
    }
    inputType = _expandIri(activeCtx, inputType, {vocab: true}, options);
  }
  const isJsonType = inputType === '@json';
```
Or if we want to go nuts, here's the cryptic version that's a challenge to indent:
```js
  const isJsonType = element[typeKey] ?
    _expandIri(
      activeCtx,
      (_isArray(element[typeKey]) && element[typeKey].length === 1) ?
        element[typeKey][0] :
        element[typeKey],
      {vocab: true},
      options) === '@json' :
    false;
```

gkellogg: Personally, I don't think it's necessary to do the _isJsonType_ bit here, and that's not what the algorithm calls for. Of course, there is no other use for it now, but it wouldn't be unreasonable to keep this as _inputType_ and do the comparison later on.

davidlehn: Ok. Shouldn't be hard to adjust later as needed.

PR #349, opened 2019-12-31
Adds support for `@included` in compaction, expansion and flattening.

gkellogg:
Code added for framing, but not enabled due to big backlog on framing in general.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,9 +37,12 @@
 
 ### Added
 - Added support for `@included` blocks
+- Skip things that have the form of a keyword, with warning.
 
 ### Fixed
 - JSON literal value handling issues.
+- Keywords may not be used as prefixes.
+
 
 ## 2.0.0 - 2019-12-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@
 
 ## 2.0.1 - 2019-12-10
 
+### Added
+- Added support for `@included` blocks
+
 ### Fixed
 - JSON literal value handling issues.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Terms of the form of an IRI must map to the same IRI.
 - Terms of the form of a relative IRI may not be used as prefixes.
 - Match spec error code "invalid context entry" vs "invalid context member".
+- Keywords may not be used as prefixes.
 
 ### Changed
 - Keep term definitions mapping to null so they may be protected.
@@ -22,6 +23,8 @@
 
 ### Added
 - Support for `"@import"`.
+- Added support for `@included` blocks
+- Skip things that have the form of a keyword, with warning.
 
 ## 2.0.2 - 2020-01-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,14 +38,8 @@
 
 ## 2.0.1 - 2019-12-10
 
-### Added
-- Added support for `@included` blocks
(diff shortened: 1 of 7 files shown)


gkellogg:
Not directly related, but it touched on some of the same code, so I added some support for keyword oddities.

CHANGELOG.md
```diff
@@ -2,8 +2,14 @@
 
 ## 2.0.1 - 2019-12-10
 
+### Added
```
davidlehn: Move all the changelog updates above 2.0.1.

lib/context.js
```diff
@@ -350,6 +350,11 @@ api.createTermDefinition = ({
       'Invalid JSON-LD syntax; keywords cannot be overridden.',
       'jsonld.SyntaxError',
       {code: 'keyword redefinition', context: localCtx, term});
+  } else if(term.match(/@[a-zA-Z]+$/)) {
```
davidlehn: Just want to find things that start with `@...` right?
```suggestion
  } else if(term.match(/^@[a-zA-Z]+$/)) {
```

davidlehn: Same with other patterns below.

davidlehn: Might be useful to move the regex to a common const at the top of the file.

lib/context.js
```diff
@@ -350,6 +350,11 @@ api.createTermDefinition = ({
...
+    // FIXME: remove logging and use a handler
+    console.warn('WARNING: terms beginning with "@" are reserved' +
```
davidlehn: Using `console.warn` here is the same issue that resulted in the `protectedMode` `warn/error` flag.  That was supposed to eventually change a better handler, but that hasn't been designed yet.  Might want another similar flag and we can fix them all at once.  The machinery to pass around those options still exists so would be easy to add.  I'm not sure what the best behavior even is.

gkellogg: I'll leave this for you to come up with a comprehensive solution. There are other areas where the spec says to generate warnings.

CHANGELOG.md
```diff
@@ -1,9 +1,14 @@
 # jsonld ChangeLog
 
-## 2.0.1 - 2019-12-10
+### Added
+- Added support for `@included` blocks
+- Skip things that have the form of a keyword, with warning.
 
 ### Fixed
 - JSON literal value handling issues.
```
davidlehn: This one line should be for 2.0.1.

lib/context.js
davidlehn: Ok.  Printing out warnings to a console can be invisible for backend code and UIs, hence the idea to have some of handler.  I haven't put time into a design for that.  A more strict mode that throws errors seems useful to me as well.

lib/context.js
```diff
@@ -437,6 +443,15 @@ api.createTermDefinition = ({
         'absolute IRI or a blank node identifier.',
         'jsonld.SyntaxError', {code: 'invalid IRI mapping', context: localCtx});
     }
+
+    if(reverse.match(KEYWORD_PATTERN)) {
+      // FIXME: remove logging and use a handler
+      console.warn('WARNING: values beginning with "@" are reserved' +
+        ' for future use and ignored', {reverse});
+      activeCtx.mappings.delete(term);
```
davidlehn: In the case of warnings (here and below), should this restore the `previousMapping` for `term` instead of deleting it?

gkellogg: Yes, it should restore before returning; I'll take care of that.

gkellogg: Fixed in 649857c33c42774640b51b83ed1e294f7d5a2b92.

PR #350, opened 2020-01-05
More support for "@type": "@none"

gkellogg:


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixed
 - JSON literal value handling issues.
+- More support for "@type": "@none"
```
davidlehn: 2.0.1 was already released, this should be above.

lib/context.js
```diff
@@ -527,7 +527,15 @@ api.createTermDefinition = ({
         {code: 'invalid type mapping', context: localCtx});
     }
 
-    if(type !== '@id' && type !== '@vocab' && type !== '@json') {
+    if((type === '@json' || type === '@none')) {
+      if(api.processingMode(activeCtx, 1.0)) {
+        throw new JsonLdError(
+          'Invalid JSON-LD syntax; an @context @type value not be ' +
```
davidlehn: Not sure exactly what this should be, perhaps:
```suggestion
          'Invalid JSON-LD syntax; an @context @type value must not be ' +
```

lib/context.js
```diff
@@ -527,7 +527,15 @@ api.createTermDefinition = ({
...
+          '#{type} in JSON-LD 1.0 mode.',
```
davidlehn: Need to use backticks and dollar sign to do interpolation.  Seems good to add quotes around the type too.
```suggestion
          `"${type}" in JSON-LD 1.0 mode.`,
```

PR #352, opened 2020-01-09
Support for `"@import"`.

gkellogg:
It might not be playing nicely with caching, please consider.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -266,6 +266,53 @@ api.process = async ({
       defined.set('@propagate', true);
     }
 
+    // handle @import
+    if('@import' in ctx) {
+      const value = ctx['@import'];
+      if(activeCtx.processingMode === 'json-ld-1.0') {
+        throw new JsonLdError(
+          'Invalid JSON-LD syntax; @import not compatible with ' +
+          activeCtx.processingMode,
+          'jsonld.SyntaxError',
+          {code: 'invalid context member', context: ctx});
+      }
+      if(!_isString(value)) {
+        throw new JsonLdError(
+          'Invalid JSON-LD syntax; @import must be a string.',
+          'jsonld.SyntaxError',
+          {code: 'invalid @import value', context: localCtx});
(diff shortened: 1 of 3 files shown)


lib/context.js
```diff
@@ -266,6 +266,53 @@ api.process = async ({
...
+      if('@import' in importCtx) {
+        throw new JsonLdError(
+          'Invalid JSON-LD syntax; imported context must not include @import.',
+          'jsonld.SyntaxError',
+          {code: 'invalid context member', context: localCtx});
+      }
+
+      // merge ctx into importCtx and replace rval with the result
+      for(const key in importCtx) {
+        if(!(key in ctx)) {
```
davidlehn: Avoid problem if importCtx has default Object property term names like 'toString':
```suggestion
        if(!ctx.hasOwnProperty(key)) {
```

lib/context.js
```diff
@@ -266,6 +266,53 @@ api.process = async ({
...
+        base: options.base
+      });
+      if(resolvedImport.length !== 1) {
+        throw new JsonLdError(
+          'Invalid JSON-LD syntax; @import must reference a single context.',
+          'jsonld.SyntaxError',
+          {code: 'invalid remote context', context: localCtx});
+      }
+      const importCtx = resolvedImport[0].document;
+      if('@import' in importCtx) {
```
davidlehn: Seems like this would need to be more involved to catch `@import` in scoped contexts too. There are not test for this.

gkellogg: Catching it was going to require adding more state to be passed down, and even though it isn't the use case, it doesn't really cause any harm if its (mis)used this way. But, that would be an issue on the spec.

davidlehn:
Does look like maybe a caching issue could happen.  I'm not sure what the situation is or if it's worth dealing with.  Maybe if you use the same resources as a context and also as an import?  Let's leave that alone until someone can figure out a test case.

PR #354, opened 2020-01-12
Miscellaneous changes

gkellogg:
A bunch of fixes that address everything other than Framing and some toRdf tests.

- Do minimal checking to see if IRIs are valid by making sure they contain no whitespace.
(804c15b93c9ff7b1c86e965a5feaa5ba0c77b3f0)
- Terms of the form of an IRI must map to the same IRI. (280a01e16bb728f4221055e102408390400f71a1)
- Terms of the form of a relative IRI may not be used as prefixes. (280a01e16bb728f4221055e102408390400f71a1)
- Keep term definitions mapping to null so they may be protected. (2613c4a26cf91d252ae7a713e666d5a90fb78ff1)
- Use localLoader for remote tests. (ff95ee28ecb99e8148c4bcbf949cf4ff7ac767f6)
- Support `rel=alternate`. (080e37445dfc39d16a80d4c56d86ce8ea392c32c)
  - Note XHR and Node loaders don't have their own tests.
- Always pass typeScopedContext to _expandObject. (f0d80f1ee1fc04a501ca7dca7d3e0dc502e3dd24)
- Allow a keyword to exist when expanding in _expandObject when the key is `@included` or `@type`. (f0d80f1ee1fc04a501ca7dca7d3e0dc502e3dd24)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@
 
 ### Changed
 - Keep term definitions mapping to null so they may be protected.
+- **NOTE**: `LINK_HEADER_REL` in `lib/constants.js` has been deprecated and
+  renamed to `LINK_HEADER_CONTEXT`. It remains for now but will be removed in a
+  future release.
 
 ## 2.0.2 - 2020-01-17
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -3,11 +3,20 @@
...
 
 ### Fixed
 - JSON literal value handling issues.
+- Do minimal checking to see if IRIs are valid by making sure they contain
+  no whitespace.
+- Terms of the form of an IRI must map to the same IRI.
+- Terms of the form of a relative IRI may not be used as prefixes.
+
+### Changed
+- Keep term definitions mapping to null so they may be protected.
```
davidlehn: Move above 2.0.1.

PR #364, opened 2020-02-19
fix typo causing missing variable

MageMasher:


lib/documentLoaders/node.js
```diff
@@ -114,13 +114,15 @@ module.exports = ({
       alternate = linkHeaders['alternate'];
       if(alternate &&
         alternate.type == 'application/ld+json' &&
+        // eslint-disable-next-line max-len
         !(res.headers['content-type'] || '').match(/^application\/(\w*\+)?json$/)) {
```
dlongley: I'd prefer a line break at `.match` to keep the style rather than a style exception.

```suggestion
        !(res.headers['content-type'] || '')
          .match(/^application\/(\w*\+)?json$/)) {
```

lib/documentLoaders/node.js
```diff
@@ -114,13 +114,15 @@ module.exports = ({
       alternate = linkHeaders['alternate'];
       if(alternate &&
         alternate.type == 'application/ld+json' &&
+        // eslint-disable-next-line max-len
```
dlongley:
```suggestion
```

PR #365, opened 2020-02-20
WIP: Framing

gkellogg:
* Remove new-embed tests, which are redundant and wrong.
- Changed framing defaults
  - `embed` to `"@once"` and warn on `"@first"` or `"@last"`.
  - `pruneBlankNodeIdentifiers` based on processingMode.
  - `omitGraph` based on processingMode.
- Top level `@graph` omitted if `omitGraph` is `true`.
- Check for invalid values of `@embed`.
- Changes in object embedding.
- Better support for graph framing.
- replaced removePreserve with cleanupPreserve and cleanupNulls
- Support default values for `@type` when framing.
- Better frame validation.


gkellogg:
This gets much of framing complete. More to do, but I may have a gap in working on it, so this can be reviewed and hopefully merged). This leaves nine failing tests.

Some tests are failing with Node6, which perhaps someone with more experience in the differences can see what might be going on.

lib/expand.js
```diff
@@ -515,7 +515,20 @@ async function _expandObject({
...
+      if(_isObject(value)) {
+        const expandedValue = {};
+        for(const [k, v] of Object.entries(value)) {
+          const kk = _expandIri(typeScopedContext, k, {vocab: true});
+          const vv = _asArray(v).map(vvv =>
+            _expandIri(typeScopedContext, vvv, {base: true, vocab: true})
+          );
+          expandedValue[kk] = vv;
+        }
+        value = expandedValue;
```
davidlehn: If using `Object.entries`, could also use `Object.fromEntries` to do this transformation:
```suggestion
        value = Object.fromEntries(Object.entries(value).map(([k, v]) => [
          _expandIri(typeScopedContext, k, {vocab: true}),
          _asArray(v).map(vv =>
￼            _expandIri(typeScopedContext, vv, {base: true, vocab: true})
￼          )
        ]));
```

gkellogg: It's hard to keep up with the evolving ES6 iterators. This is, indeed, cleaner.

gkellogg:
It looks like the Node 6 failure is due to `Object.entries` not being defined. There is a suggestion for a [PolyFill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries#Polyfill) on the [MDN entry for Object.entries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries).

Where would be the best place to apply this?

davidlehn:
Short answer is to hack it to work for now.  core-js for Object.entries, and a shim for Object.fromEntries.  I'll try adding it.  Was trying to avoid this nonsense in jsonld.js 2.x.  Once 1.1 support is mostly or all done I think we'll get radical with changes for a 3.x release.  That'll include updating to core-js 3, which has Object.fromEntries, and just dropping node6 support and see if anyone notices.

lib/frame.js
```diff
@@ -110,20 +110,53 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
 
     // if embed is @never or if a circular reference would be created by an
     // embed, the subject cannot be embedded, just add the reference;
     // note that a circular reference won't occur when the embed flag is
     // `@link` as the above check will short-circuit before reaching this point
-    if(flags.embed === '@never' ||
-      _createsCircularReference(subject, state.graph, state.subjectStack)) {
+    if(state.embedded &&
+      (flags.embed === '@never' ||
+      _createsCircularReference(subject, state.graph, state.subjectStack)))
```
dlongley:
```suggestion
      _createsCircularReference(subject, state.graph, state.subjectStack))) {
```

lib/frame.js
```diff
@@ -110,20 +110,53 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
+    {
```
dlongley:
```suggestion
```

lib/frame.js
```diff
@@ -110,20 +110,53 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
+      _addFrameOutput(parent, property, output);
+      continue;
+    }
+
+    // if only the first (or once) should be embedded
+    if(state.embedded &&
+       (flags.embed == '@first' || flags.embed == '@once') &&
+       state.uniqueEmbeds[state.graph].hasOwnProperty(id))
```
dlongley:
```suggestion
       state.uniqueEmbeds[state.graph].hasOwnProperty(id)) {
```

lib/frame.js
```diff
@@ -110,20 +110,53 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
+    {
```
dlongley:
```suggestion
```

lib/frame.js
```diff
@@ -134,43 +167,42 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
-        recurse = !(id === '@merged' || id === '@default');
       }
 
       if(recurse) {
         state.graphStack.push(state.graph);
-        state.graph = id;
         // recurse into graph
         api.frame(
-          state,
+          Object.assign({}, state, {graph: id, embedded: false}),
```
dlongley:
```suggestion
          {...state, graph: id, embedded: false},
```

lib/frame.js
```diff
@@ -134,43 +167,42 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
-        state.graph = state.graphStack.pop;
+        state.graphStack.pop;
       }
     }
 
     // if frame has @included, recurse over its sub-frame
     if('@included' in frame) {
       api.frame(
-        state,
+        Object.assign({}, state, {embedded: false}),
```
dlongley:
```suggestion
        {...state, embedded: false}),
```

lib/frame.js
```diff
@@ -198,36 +230,37 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
-            o = src[n];
-            if(graphTypes.isSubjectReference(o)) {
-              const subframe = (prop in frame ?
-                frame[prop][0]['@list'] : _createImplicitFrame(flags));
+          for(const oo of src) {
+            if(graphTypes.isSubjectReference(oo)) {
               // recurse into subject reference
-              api.frame(state, [o['@id']], subframe, list, '@list');
+              api.frame(
+                Object.assign({}, state, {embedded: true}),
```
dlongley:
```suggestion
                {...state, embedded: true}),
```

lib/frame.js
```diff
@@ -198,36 +230,37 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
           }
-          continue;
-        }
-
-        if(graphTypes.isSubjectReference(o)) {
+        } else if(graphTypes.isSubjectReference(o)) {
           // recurse into subject reference
-          api.frame(state, [o['@id']], subframe, output, prop);
+          api.frame(
+            Object.assign({}, state, {embedded: true}),
```
dlongley:
```suggestion
            {...state, embedded: true}),
```

lib/frame.js
```diff
@@ -238,7 +271,14 @@ api.frame = (state, subjects, frame, parent, property = null) => {
     // handle defaults
     for(const prop of Object.keys(frame).sort()) {
       // skip keywords
-      if(isKeyword(prop)) {
+      if(prop === '@type') {
+        if(!types.isObject(frame[prop][0]) ||
+           !frame[prop][0].hasOwnProperty('@default'))
```
dlongley:
```suggestion
           !frame[prop][0].hasOwnProperty('@default')) {
```

lib/frame.js
```diff
@@ -238,7 +271,14 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
+        {
```
dlongley:
```suggestion
```

lib/frame.js
```diff
@@ -260,21 +300,20 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
+      for(const subject of Object.keys(state.subjects)) {
+        const nodeValues =
+          util.getValues(state.subjects[subject], reverseProp);
+        if(nodeValues.some(v => v['@id'] === id)) {
+          // node has property referencing this subject, recurse
+          output['@reverse'] = output['@reverse'] || {};
+          util.addValue(
+            output['@reverse'], reverseProp, [], {propertyIsArray: true});
+          api.frame(
+            Object.assign({}, state, {embedded: true}),
```
dlongley:
```suggestion
            {...state, embedded: true}),
```

lib/frame.js
```diff
@@ -346,11 +429,15 @@ function _getFrameFlag(frame, options, name) {
...
     // false => "@never"
     if(rval === true) {
-      rval = '@last';
+      rval = '@once';
     } else if(rval === false) {
       rval = '@never';
-    } else if(rval !== '@always' && rval !== '@never' && rval !== '@link') {
-      rval = '@last';
+    } else if(rval !== '@always' && rval !== '@never' && rval !== '@link' &&
+              rval !== '@first' && rval !== '@last' && rval !== '@once')
```
dlongley:
```suggestion
      rval !== '@first' && rval !== '@last' && rval !== '@once') {
```

lib/frame.js
```diff
@@ -346,11 +429,15 @@ function _getFrameFlag(frame, options, name) {
...
+    {
```
dlongley:
```suggestion
```

lib/frame.js
```diff
@@ -367,6 +454,32 @@ function _validateFrame(frame) {
       'Invalid JSON-LD syntax; a JSON-LD frame must be a single object.',
       'jsonld.SyntaxError', {frame});
   }
+
+  if(frame[0].hasOwnProperty('@id')) {
+    for(const id of util.asArray(frame[0]['@id'])) {
+      // @id must be wildcard or an IRI
+      if(!(types.isObject(id) || url.isAbsolute(id)) ||
+         (types.isString(id) && id.indexOf('_:') === 0))
```
dlongley:
```suggestion
         (types.isString(id) && id.indexOf('_:') === 0)) {
```

lib/frame.js
```diff
@@ -367,6 +454,32 @@ function _validateFrame(frame) {
...
+      {
```
dlongley:
```suggestion
```

lib/frame.js
```diff
@@ -367,6 +454,32 @@ function _validateFrame(frame) {
...
+          'jsonld.SyntaxError', {code: 'invalid frame', frame});
+      }
+    }
+  }
+
+  if(frame[0].hasOwnProperty('@type')) {
+    for(const type of util.asArray(frame[0]['@type'])) {
+      // @id must be wildcard or an IRI
+      if(!(types.isObject(type) || url.isAbsolute(type)) ||
+         (types.isString(type) && type.indexOf('_:') === 0))
```
dlongley:
```suggestion
         (types.isString(type) && type.indexOf('_:') === 0)) {
```

lib/frame.js
```diff
@@ -367,6 +454,32 @@ function _validateFrame(frame) {
...
+      {
```
dlongley:
```suggestion
```

lib/frame.js
```diff
@@ -419,105 +532,106 @@ function _filterSubject(state, subject, frame, flags) {
...
+      } else if(types.isObject(thisFrame)) { // XXX only framing keywords
+        // node matches if values is not empty and the value of property in frame
+        // is wildcard
+        matchThis = nodeValues.length > 0;
+      } else {
+        if(graphTypes.isValue(thisFrame)) {
+          // match on any matching value
+          matchThis = nodeValues.some(nv => _valueMatch(thisFrame, nv));
+        } else if(graphTypes.isSubject(thisFrame) ||
+          graphTypes.isSubjectReference(thisFrame))
```
dlongley:
```suggestion
          graphTypes.isSubjectReference(thisFrame)) {
```

lib/frame.js
```diff
@@ -419,105 +532,106 @@ function _filterSubject(state, subject, frame, flags) {
...
+        {
```
dlongley:
```suggestion
```

lib/util.js
```diff
@@ -147,31 +147,33 @@ api.parseLinkHeader = header => {
...
+    return;
+  } else if(isFrame && types.isObject(v)) {
+    switch(Object.keys(v).length) {
+      case 0:
+        // empty object is wildcard
+        return;
+      case 1:
+        // default entry is all strings
+        if('@default' in v &&
+           api.asArray(v['@default']).every(vv => types.isString(vv)))
```
dlongley:
```suggestion
           api.asArray(v['@default']).every(vv => types.isString(vv))) {
```

lib/util.js
```diff
@@ -147,31 +147,33 @@ api.parseLinkHeader = header => {
...
-    // must contain only strings
-    isValid = true;
-    for(let i = 0; i < v.length; ++i) {
-      if(!(types.isString(v[i]))) {
-        isValid = false;
-        break;
-      }
+  if(types.isArray(v) && v.every(vv => types.isString(vv))) {
+    return;
+  } else if(isFrame && types.isObject(v)) {
```
dlongley: Can just be `if` since there's an early return.

lib/util.js
```diff
@@ -147,31 +147,33 @@ api.parseLinkHeader = header => {
...
+    switch(Object.keys(v).length) {
+      case 0:
+        // empty object is wildcard
+        return;
+      case 1:
+        // default entry is all strings
+        if('@default' in v &&
+           api.asArray(v['@default']).every(vv => types.isString(vv)))
+        {
```
dlongley:
```suggestion
```

lib/frame.js
```diff
@@ -419,105 +532,106 @@ function _filterSubject(state, subject, frame, flags) {
...
-              state, listValue, lv, flags));
+
+      if(thisFrame === undefined) {
+        // node does not match if values is not empty and the value of property
+        // in frame is match none.
+        if(nodeValues.length > 0) {
+          return false;
+        }
+        matchThis = true;
+      } else if(types.isObject(thisFrame)) { // XXX only framing keywords
```
davidlehn: Does this "XXX" comment need to be addressed?

davidlehn:
@dlongley Re linting, we do have eslint config here, and a "lint" command, but our rules didn't check for 'brace-style'.  Added a PR to consider for that:
https://github.com/digitalbazaar/eslint-config-digitalbazaar/pull/31

lib/frame.js
```diff
@@ -134,43 +165,42 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
       if(recurse) {
         state.graphStack.push(state.graph);
-        state.graph = id;
         // recurse into graph
         api.frame(
-          state,
+          {...state, graph: id, embedded: false},
           Object.keys(state.graphMap[id]).sort(), [subframe], output, '@graph');
-        state.graph = state.graphStack.pop;
+        state.graphStack.pop;
```
davidlehn: Should this be calling pop?  Should it have been calling it before?

lib/frame.js
```diff
@@ -134,43 +165,42 @@ api.frame = (state, subjects, frame, parent, property = null) => {
...
+
     // push matching subject onto stack to enable circular embed checks
     state.subjectStack.push({subject, graph: state.graph});
 
     // subject is also the name of a graph
     if(id in state.graphMap) {
       let recurse = false;
       let subframe = null;
-      if(!('@graph' in frame)) {
+      if(!frame.hasOwnProperty('@graph')) {
```
davidlehn: We don't really need to do `hasOwnProperty` for static strings.  Really more of an issue when it's a dynamic string that could conflict with an inherited Object property.  Probably doesn't matter, but the `in` syntax looks nicer when it's ok to use it.

lib/frame.js
gkellogg: It was in my code, and it could relate to some of the yet-to-be fixed issues.

lib/frame.js
gkellogg: We pushed on to the stack before recursing, so need to pop it back off when done. We don't need to save that, since `state.graph` was set just for the call.

lib/frame.js
gkellogg:
```suggestion
      if(!('@graph' in frame)) {
```

lib/frame.js
davidlehn: Before it was setting `state.graph` to the pop function.  Now it's just doing nothing.  I assume this should be `pop()`, but it doesn't seem to matter either way in the test suite.

gkellogg: No, you're right, should be `pop()` (Ruby-isms :) ). It doesn't matter for the tests, but it might in production.

I'll test before committing an update

davidlehn: Fixing and filing an issue about it.

PR #366, opened 2020-02-23
Add event handler

davidlehn:
- An event handler option seemed like a good approach for a way to allow custom handling of warnings (to display in UIs or throw errors or whatever).
- I was thinking a "strict mode" handler could be available that would turn warnings into errors.
- The event approach with chainable handlers also turned out to work well for the case of capturing and replaying events for cached processed contexts.
- I imagine other events beyond warnings might be useful.
- Open to ideas for alternate approaches.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -20,6 +20,10 @@ const {
 } = require('./url');
 
 const {
+  handleEvent: _handleEvent
+} = require('./events');
+
+const {
   asArray: _asArray,
   compareShortestLeast: _compareShortestLeast
 } = require('./util');
@@ -60,6 +64,23 @@ api.process = async ({
   if(ctxs.length === 0) {
     return activeCtx;
   }
+
+  // event handler for capturing events to replay when using a cached context
+  const events = [];
+  const handleEvent = [
+    ({event, next}) => {
+      events.push(event);
+      next();
+    }
+  ];
+  // chain to original handler
+  if(options.handleEvent) {
+    handleEvent.push(options.handleEvent);
+  }
+  // store original options to use when replaying events
+  const originalOptions = options;
+  // shallow clone options with custom event handler
+  options = Object.assign({}, options, {handleEvent});
 
   // resolve contexts
   const resolved = await options.contextResolver.resolve({
@@ -112,7 +133,12 @@ api.process = async ({
     // get processed context from cache if available
     const processed = resolvedContext.getProcessed(activeCtx);
     if(processed) {
-      rval = activeCtx = processed;
+      // replay events with original non-capturing options
+      for(const event of processed.events) {
+        _handleEvent({event, options: originalOptions});
+      }
+
+      rval = activeCtx = processed.context;
       continue;
     }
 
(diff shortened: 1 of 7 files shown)


lib/context.js
```diff
@@ -517,9 +526,19 @@ api.createTermDefinition = ({
     }
 
     if(reverse.match(KEYWORD_PATTERN)) {
```
davidlehn: What's an example where this would match?  Looks like the above checks wouldn't allow `{"@reverse": "@foo"}` or similar to get this far.

gkellogg (approved):

lib/context.js
gkellogg: Previous checks don't specifically look at `@reverse`, except to verify that it's a string.

davidlehn:
Some rambling thoughts on this patch:
- "handleEvent" params should be renamed "eventHandler".
- I used "code" to align with error codes.  Pondering how to best handle "code", "level", and maybe "type" concepts.
- I started adding events for all the places where expansionMap/compactionMap skip something. 
 Unsure how best to go about that as their functionality starts to overlap with events.  The spec doesn't say dropping unknown data is a "warning", so I'm not sure what to call it to distinguish the two flows.  "debug"?  Maybe levels are a bad idea and it should be more like `"defaultAction": "warning"`?  May need to fiddle with this to avoid too much output, yet be useful for debug tools.

lib/context.js
davidlehn: It checks it's a string, `_expandIris` it, then `_isAbsoluteIri` checks that.  Seems like a keyword like string wouldn't get beyond those steps.  Is there an example where it would?

gkellogg: Just created https://github.com/w3c/json-ld-api/pull/389, as the order of checks was inverted.

lib/context.js
```diff
@@ -61,6 +65,23 @@ api.process = async ({
...
+    }
+  ];
+  // chain to original handler
+  if(options.handleEvent) {
+    handleEvent.push(options.handleEvent);
+  }
+  // store original options to use when replaying events
+  const originalOptions = options;
+  // shallow clone options with custom event handler
+  options = Object.assign({}, options, {handleEvent});
```
dlongley:
```suggestion
  options = {...options, handleEvent});
```

dlongley (approved):
Thanks for taking care of this @davidlehn!

I agree with your latest comments ... and I would also ask, any reason why the event handler stuff can't support async? We could add `await` code, etc. -- and it would only get called whenever event handlers were actually installed. My understanding is that we made the processing functions async so this should be possible... but I also feel like we discussed this already and you said it couldn't/shouldn't be done for some reason, but I can't recall what the issue was.

davidlehn:
Some of the events happen in createTermDefinition.  If event handlers were to be async, we'd have to make that async, and callers like _expandIri async, and maybe others up the stack.  I was afraid that might have performance effects, but I didn't check.  Another case where the in-progress benchmarking code needs to be improved so we can test such changes.  I figured for version 1 the handlers could be sync and if we change to async/await in the future, they'll still work.

PR #369, opened 2020-02-27
Handle term definition on `@type` with empty map.

davidlehn:
Fixes expand#tec02 & toRdf#tec02.

@gkellogg "keyword redefinition" was not what I expected for this type of error.  Is that more appropriate than the more generic "invalid term definition"?


gkellogg:
> Fixes expand#tec02 & toRdf#tec02.
> 
> @gkellogg "keyword redefinition" was not what I expected for this type of error. Is that more appropriate than the more generic "invalid term definition"?

It's "keyword redefinition" as that is the case for any attempt to change a keyword, other than specific carve-outs for `@type`. One could argue for "invalid term definition", but it's not a term, and it's a bit late to change the spec.

PR #370, opened 2020-02-27
Fix term mapping of `@reverse`.

davidlehn:
Please check.  I'm making tests pass without proper understanding of how this should be working.  Should there be some warning here or is the "ignore" behavior itself important?

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -506,18 +506,7 @@ api.createTermDefinition = ({
         'jsonld.SyntaxError', {code: 'invalid IRI mapping', context: localCtx});
     }
 
-    // expand and add @id mapping
-    const id = _expandIri(
-      activeCtx, reverse, {vocab: true, base: false}, localCtx, defined,
-      options);
-    if(!_isAbsoluteIri(id)) {
-      throw new JsonLdError(
-        'Invalid JSON-LD syntax; a @context @reverse value must be an ' +
-        'absolute IRI or a blank node identifier.',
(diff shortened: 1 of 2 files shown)


gkellogg:
> Please check. I'm making tests pass without proper understanding of how this should be working. Should there be some warning here or is the "ignore" behavior itself important?

The "ignore" behavior is important, as it affects processing flow. You should also generate a warning.

The bug I had in my code was I checked after expanding it as an IRI, which led to a failure, which is not right. With the ignore/warning, you drop out of the create term definition for that term and a term doesn't get created.

lib/context.js
```diff
@@ -505,6 +505,10 @@ api.createTermDefinition = ({
         'jsonld.SyntaxError', {code: 'invalid IRI mapping', context: localCtx});
     }
 
+    if(!api.isKeyword(reverse) && reverse.match(KEYWORD_PATTERN)) {
+      return;
```
gkellogg: Should issue a warning here.

lib/context.js
```diff
@@ -516,18 +520,6 @@ api.createTermDefinition = ({
...
-      console.warn('WARNING: values beginning with "@" are reserved' +
-        ' for future use and ignored', {reverse});
-      if(previousMapping) {
-        activeCtx.mappings.set(term, previousMapping);
-      } else {
-        activeCtx.mappings.delete(term);
-      }
-      return;
-    }
-
```
gkellogg: Probably need to continue to revert to a previousMapping if ignoring.

PR #373, opened 2020-03-02
Fix for: Framing does not respect specified value #300

### LINKED ISSUE #300: Framing does not respect specified value
**Expected Behaviour**: Given a document with 2 persons "Jane" and "John" and a frame for `"givenName": "John"` I expect the result to contain only John, but not Jane.

**Actual Behaviour**: The result contains both persons, just that Janes givenName is set to `null`

I have set up the [exact scenario in the playground here](https://json-ld.org/playground/#startTab=tab-framed&json-ld=%7B%22%40context%22%3A%7B%22%40version%22%3A1.1%2C%22%40vocab%22%3A%22https%3A%2F%2Fschema.org%2F%22%7D%2C%22%40graph%22%3A%5B%7B%22%40id%22%3A%221%22%2C%22%40type%22%3A%22Person%22%2C%22name%22%3A%22John%20Doe%22%2C%22givenName%22%3A%22John%22%2C%22familyName%22%3A%22Doe%22%7D%2C%7B%22%40id%22%3A%222%22%2C%22%40type%22%3A%22Person%22%2C%22name%22%3A%22Jane%20Doe%22%2C%22givenName%22%3A%22Jane%22%7D%5D%7D&frame=%7B%22%40context%22%3A%7B%22%40version%22%3A1.1%2C%22%40vocab%22%3A%22https%3A%2F%2Fschema.org%2F%22%7D%2C%22givenName%22%3A%22John%22%7D).

The expected result there would be:

```
{
  "@context": {
    "@version": 1.1,
    "@vocab": "https://schema.org/"
  },
  "@id": "1",
  "@type": "Person",
  "name": "John Doe",
  "givenName": "John",
  "familyName": "Doe"
}
```

**Affected versions**
- jsonld playground as  of 2019-03-04
- jsonld npm package version 1.5.4
- works *as expected* in the [RDF Distiller](http://rdf.greggkellogg.net/distiller?command=serialize)

michaelcpuckett:
This fixes #300 to match behavior in RDF Distiller.

> Expected Behaviour: Given a document with 2 persons "Jane" and "John" and a frame for "givenName": "John" I expect the result to contain only John, but not Jane.
> 
> Actual Behaviour: The result contains both persons, just that Janes givenName is set to null

This removes the initial check on the node to see if it has any properties and falls through to check the node to see if the values match.

I added a test based on the issue.

### DIFF
--- a/lib/frame.js
+++ b/lib/frame.js
@@ -596,10 +596,6 @@ function _filterSubject(state, subject, frame, flags) {
           return false;
         }
         matchThis = true;
-      } else if(types.isObject(thisFrame)) { // XXX only framing keywords
-        // node matches if values is not empty and the value of property in
-        // frame is wildcard
-        matchThis = nodeValues.length > 0;
       } else {
         if(graphTypes.isValue(thisFrame)) {
           // match on any matching value
--- a/tests/misc.js
+++ b/tests/misc.js
@@ -460,6 +460,48 @@ describe('js keywords', () => {
     const e = await jsonld.frame(d, frame);
     assert.deepStrictEqual(e, ex);
   });
+
+  it('frame js match on value', async () => {
+    const d =
+{
+  "@context": {
+    "@vocab": "https://schema.org/"
+  },
+  "@graph": [
+    {
+      "@id": "JOHN",
+      "@type": "Person",
+      "givenName": "John"
+    },
+    {
+      "@id": "JANE",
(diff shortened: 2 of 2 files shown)


michaelcpuckett:
I see this is failing some tests. I didn't run fetch-test-suites first. I'll keep looking into it, but closing for now.

PR #374, opened 2020-03-02
Fix for: Framing does not respect specified value #300

### LINKED ISSUE #300: Framing does not respect specified value

michaelcpuckett:
This fixes #300 to match behavior in RDF Distiller.

> Expected Behaviour: Given a document with 2 persons "Jane" and "John" and a frame for "givenName": "John" I expect the result to contain only John, but not Jane.

> Actual Behaviour: The result contains both persons, just that Janes givenName is set to null

This moves the initial check on the node to see if it has any properties so that value matches can be checked.

I added a test based on the issue.

(Originally opened #373 but spotted an issue after checking the test-suites -- Now all tests are passing locally.)

### DIFF
--- a/lib/frame.js
+++ b/lib/frame.js
@@ -596,19 +596,8 @@ function _filterSubject(state, subject, frame, flags) {
           return false;
         }
         matchThis = true;
-      } else if(types.isObject(thisFrame)) { // XXX only framing keywords
-        // node matches if values is not empty and the value of property in
-        // frame is wildcard
-        matchThis = nodeValues.length > 0;
       } else {
-        if(graphTypes.isValue(thisFrame)) {
-          // match on any matching value
-          matchThis = nodeValues.some(nv => _valueMatch(thisFrame, nv));
-        } else if(graphTypes.isSubject(thisFrame) ||
-          graphTypes.isSubjectReference(thisFrame)) {
-          matchThis =
-            nodeValues.some(nv => _nodeMatch(state, thisFrame, nv, flags));
-        } else if(graphTypes.isList(thisFrame)) {
+        if(graphTypes.isList(thisFrame)) {
           const listValue = thisFrame['@list'][0];
           if(graphTypes.isList(nodeValues[0])) {
             const nodeListValues = nodeValues[0]['@list'];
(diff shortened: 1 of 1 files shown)


michaelcpuckett:
This is passing the new test, but a more complicated test is failing locally. I'll leave this open while I look further.

davidlehn:
A test like this should be added to the main test suite too.

michaelcpuckett:
OK, it's working now.

I moved the 3 tests to here:

https://github.com/w3c/json-ld-framing/pull/100

I emailed the group regarding adding the tests.

lib/frame.js
```diff
@@ -596,35 +596,30 @@ function _filterSubject(state, subject, frame, flags) {
...
       } else {
-        if(graphTypes.isValue(thisFrame)) {
-          // match on any matching value
-          matchThis = nodeValues.some(nv => _valueMatch(thisFrame, nv));
-        } else if(graphTypes.isSubject(thisFrame) ||
-          graphTypes.isSubjectReference(thisFrame)) {
-          matchThis =
-            nodeValues.some(nv => _nodeMatch(state, thisFrame, nv, flags));
-        } else if(graphTypes.isList(thisFrame)) {
+        if (graphTypes.isList(thisFrame)) {
```
davidlehn: Please run "npm run lint".  Style we use omits space after `if`.

michaelcpuckett: OK, should be resolved now.

gkellogg:
You can remove the following lines from test-common:

```
        /frame-manifest.jsonld#tin01$/,
        /frame-manifest.jsonld#tin02$/,
```

This still leaves 2 list tests and one included test, which might have a related fix.

gkellogg:
It ended up that tin03 was failing because framing was order dependent. I updated the test to simplify, so tin03 passes too and the following can be removed from test-common.js:

```
        // included
        /frame-manifest.jsonld#tin01$/,
        /frame-manifest.jsonld#tin02$/,
        /frame-manifest.jsonld#tin03$/,
```

davidlehn:
Thanks!

PR #376, opened 2020-03-04
Default frame for lists.

gkellogg:
The code previously would get an invalid frame, if there was a frame object, but it didn't have an `@list` entry.

### DIFF
--- a/lib/frame.js
+++ b/lib/frame.js
@@ -234,9 +234,9 @@ api.frame = (state, subjects, frame, parent, property = null) => {
 
         // recurse into list
         if(graphTypes.isList(o)) {
-          const subframe = (frame[prop] && types.isObject(frame[prop][0]) ?
-            (frame[prop][0]['@list'] || _createImplicitFrame(flags)) :
-            _createImplicitFrame(flags));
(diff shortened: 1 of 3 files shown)


gkellogg:
Test failures should go away when PR #374 is merged.

lib/frame.js
```diff
@@ -235,7 +235,8 @@ api.frame = (state, subjects, frame, parent, property = null) => {
         // recurse into list
         if(graphTypes.isList(o)) {
           const subframe = (frame[prop] && types.isObject(frame[prop][0]) ?
-            frame[prop][0]['@list'] : _createImplicitFrame(flags));
+            (frame[prop][0]['@list'] || _createImplicitFrame(flags)) :
+            _createImplicitFrame(flags));
```
davidlehn: Logic can be compacted, shift `@list` check in to the conditional:
```suggestion
          const subframe = (frame[prop] && types.isObject(frame[prop][0] && frame[prop][0]['@list']) ?
￼            frame[prop][0]['@list'] : _createImplicitFrame(flags));
```

gkellogg: Fixed in faaaf6a01db684ef2871343c989a0d97849e3a5b.

Issue #371, opened 2020-02-28
frame.js state.graphStack is not used.


davidlehn:
`lib/frame.js` tracks graphs with `state.graphStack` but it doesn't look like they are used.

@gkellogg Was there a plan for this?  Pending tests that would need it?  Maybe something that didn't make it over from ruby?

It might be a good point for a debugging hook.

gkellogg:
Perhaps we should add tests that would flesh out the need for this.

Intuitively, if pushing onto a shared stack when processing a named graph, you would need to revert to the previous state after processing, but a test that could fail would be useful. Or, if we can't come up with one, perhaps it's not really necessary after all.

gkellogg:
See https://github.com/w3c/json-ld-framing/pull/97, I now believe `graphStack` serves no purpose.

davidlehn:
Closed in https://github.com/digitalbazaar/jsonld.js/pull/377.

PR #377, opened 2020-03-06
Remove framing `graphStack` code.

davidlehn:
- Code was unused and has been removed from the framing spec.


davidlehn:
Closes https://github.com/digitalbazaar/jsonld.js/issues/371.

Issue #358, opened 2020-01-27
`@included` behaves incorrectly if `@context` follows entry


hsolbrig:
See: http://tinyurl.com/r8chydx

If the `@context` entry precedes the `@included`, behavior is as expected.  If at the end, `@included` maps to `http://example.org/@included`

gkellogg:
Full support for `@included` is in https://github.com/digitalbazaar/jsonld.js/pull/349.

hsolbrig:
The issue seems to have been fixed in the above example, but https://tinyurl.com/ura2qxv 
still has it.

davidlehn:
A newer jsonld.js was released and the playground updated.  Did that solve this issue?

hsolbrig:
Unfortunately, I'm getting a hang when I use the above link.

davidlehn:
Yup.  I was hoping no one would notice. ;-)  Locking the whole tab up, which is strange.  Maybe something recursive in those contexts?  Will have to investigate.

hsolbrig:
Ya - the contexts are recursive as all getout -- I thought that was ok...

davidlehn:
In Node.js it first flips out on https://fhircat.org/fhir/contexts/r5/medicationadministration.context.jsonld.

No special encoding was specified:
```
$ curl -I https://fhircat.org/fhir/contexts/r5/medicationadministration.context.jsonld
[...]
Content-Type: application/ld+json
[...]
```

It has BOM bytes:
```
$ file medicationadministration.context.jsonld 
medicationadministration.context.jsonld: UTF-8 Unicode (with BOM) text
```
```
$ curl -O https://fhircat.org/fhir/contexts/r5/medicationadministration.context.jsonld
$ xxd -l 16 medicationadministration.context.jsonld
00000000: efbb bf7b 0a20 2022 4063 6f6e 7465 7874  ...{.  "@context
```

`JSON.parse` isn't liking that in node:
```
  [...]
  name: 'jsonld.InvalidUrl',
  details: {
    code: 'loading remote context failed',
    url: 'https://fhircat.org/fhir/contexts/r5/medicationadministration.context.jsonld',   
    cause: SyntaxError: Unexpected token ﻿ in JSON at position 0
        at JSON.parse (<anonymous>)
        at ContextResolver._fetchContext (.../jsonld.js/lib/ContextResolver.js:159:24)
  [...]
```

[...]

PR #380, opened 2020-03-12
Remove BOM character from context if there is one

jiaola:
Indirectly related to #358. When the context file contains a BOM character, there's an error: 

    cause: SyntaxError: Unexpected token ﻿ in JSON at position 0
        at JSON.parse (<anonymous>)
        at ContextResolver._fetchContext (.../jsonld.js/lib/ContextResolver.js:159:24)

This PR fixes it.

### DIFF
--- a/lib/ContextResolver.js
+++ b/lib/ContextResolver.js
@@ -153,6 +153,9 @@ module.exports = class ContextResolver {
       context = remoteDoc.document || null;
       // parse string context as JSON
       if(_isString(context)) {
+        if (context.charCodeAt(0) === 0xFEFF) {
+          context = context.slice(1);
+        }
         context = JSON.parse(context);
       }
     } catch(e) {


lib/ContextResolver.js
```diff
@@ -153,6 +153,9 @@ module.exports = class ContextResolver {
       context = remoteDoc.document || null;
       // parse string context as JSON
       if(_isString(context)) {
+        if (context.charCodeAt(0) === 0xFEFF) {
+          context = context.slice(1);
```
dlongley: This seems to be happening at the wrong layer. I would expect, if anything is to handle this, it would be a document loader that would strip it from the bytes it receives.

gkellogg:
Why is this something a client needs to do at all? From RFC7159, it’s illegal for, JSON data to include a BOM.

https://tools.ietf.org/html/rfc7159#section-8.1

jiaola:
Good point! Let me talk with Harold Solbrig about this. I wasn't aware that BOM is illegal in JSON. I'll close this PR.

hsolbrig:
We'll fix the emitter.  

That said, it would be handy emit a warning or error that says something a we bit more descriptive than "bad character"? My hunch is that we're not the only ones that will encounter this problem and I suspect that a goodly number of JSON users may never have heard of byte-ordering marks.  

Is there a site out there that tells folks how to fix this problem in their language of choice?  If not, should we add one to the best practices site?

gkellogg:
There’s a good page here: https://www.w3.org/International/questions/qa-byte-order-mark

The RFC I cited shows how it’s inappropriate for application/json, and this JSON-LD

davidlehn:
On the playground the BOM files were getting parsed.  So the XHR loader or browser XHR itself is handling it.  Node loader not so much.  A "bad character" error isn't wrong, but worth considering also doing a BOM check and making that particular error more clear.  Also a future documentLoader implementation could have a BOM check/ignore option for those who want it.

PR #381, opened 2020-03-13
Context overflow.

davidlehn:
- Add support for context overflow tests: https://github.com/w3c/json-ld-api/pull/416
- Attempt to support older 1.0 error code.  Is this needed?  (It required passing some activeCtx around to get to processing mode)
- Note: The MAX value only detects *depth* of recursion, not total number of remote contexts.  This may be inline with the former "recursion" error code, but seems the "context overflow" should now check total remote contexts loaded (which would also handle depth).  Unsure how to change the code to track that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 ### Changed
 - Better support for using a processed context for `null` and caching
   `@import`.
+- Don't set `@base` in initial context and don't resolve a relative IRI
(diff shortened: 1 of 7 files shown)


gkellogg:
@davidlehn it would be nice to merge this, as other branches will fail until rebased on this work. (of course, it would be nice to get some of those merged too).

PR #384, opened 2020-03-27
Fixed `prependBase` 

gkellogg:
to start path with a '/' for a zero length path if there is an authority in base.

For expand#t0129 and toRdf#te129.


gkellogg:
Needs to be rebased after PR #381 is merged.

PR #390, opened 2020-04-15
Add various list + base tests.

davidlehn:
- WIP.
- Started as a test to debug a N-Quads serialization bug.
- Also checking some list + base behavior.

### DIFF
--- a/tests/misc.js
+++ b/tests/misc.js
@@ -159,6 +159,150 @@ describe('other toRDF tests', () => {
       done();
     });
   });
+
+  it.only('should handle list good @context @base', async () => {
+    const doc = {
+      "@context": {
+        "@base": "ex:",
+        "list": {
+          "@id": "foo:bar",
+          "@container": "@list",
+          "@type": "@id"
+        }
+      },
+      "list": [
+        "test"
+      ]
+    };
+    const rdf = await jsonld.toRDF(doc, {
+      format: 'application/n-quads'
+    });
+    assert.equal(
+      rdf,
+      '_:b0 <foo:bar> _:b1 .\n' +
+      '_:b1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#first> <ex:test> .\n' +
+      '_:b1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#rest> <http://www.w3.org/1999/02/22-rdf-syntax-ns#nil> .\n');
+  });
+
+  it.only('should handle list bad @context @base', async () => {
+    const doc = {
+      "@context": {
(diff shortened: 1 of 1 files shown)


tests/misc.js
```diff
@@ -159,6 +159,150 @@ describe('other toRDF tests', () => {
...
+      ]
+    };
+    const rdf = await jsonld.toRDF(doc, {
+      base: null,
+      format: 'application/n-quads'
+    });
+    assert.equal(
+      rdf,
+      '_:b0 <foo:bar> _:b1 .\n' +
+      //'_:b1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#first> <test> .\n' +
```
gkellogg, line 255: This, and others, might be controversial. Certainly, the spec says that the base option overrides the input document's IRI, but this would correspond to the [RFC3986 5.1.2 Base URI from the Encapsulating Entity](https://tools.ietf.org/html/rfc3986#section-5.1.2), and following their hierarchy, an empty value would typically be made absolute by resaving against the Retrieval URI or the default base URI. It could be that it's just that the spec wording is muddy. Does it make sense to set the base option to anything other than an absolute URL?

gkellogg: So, arguably, the API should ignore a base option which is not a valid IRI (meaning absolute). The option is described as setting the `base IRI`, and `base IRI` is described as an `IRI`, meaning absolute:

> The `base IRI` is an [IRI](https://tools.ietf.org/html/rfc3987#section-2) established in the context, or is based on the JSON-LD document location. The `base IRI` is used to turn relative IRI references into IRIs.

We don't describe what to do for invalid options, so it's a bit of a gray area.

tests/misc.js
```diff
@@ -159,6 +159,150 @@ describe('other toRDF tests', () => {
...
+      '_:b0 <foo:bar> _:b1 .\n' +
+      '_:b1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#first> <ex:test> .\n' +
+      '_:b1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#rest> <http://www.w3.org/1999/02/22-rdf-syntax-ns#nil> .\n');
+  });
+
+  it.only('should handle list bad @context @base', async () => {
+    const doc = {
+      "@context": {
+        "@base": "::",
+        "list": {
```
gkellogg, line 191: This isn't invalid, as ':' is a valid char, which is a component of segment-nz. When expanding relative to the document base, it becomes `https://w3c.github.io/json-ld-api/tests/toRdf/::`, which is valid.

I'll pick something else.

Issue #394, opened 2020-05-13
Minimal scoped context affects term selection with index container


gkellogg:
From https://github.com/w3c/wot-thing-description/issues/894

Compacting the following:

```json
{
  "https://www.w3.org/2019/wot/td#hasPropertyAffordance": [
    {
      "http://purl.org/dc/terms/title": "Switch"
    }
  ],
  "http://purl.org/dc/terms/title": "46203"
}
```

with the following context:

```json
{
  "@context": {
    "td": "https://www.w3.org/2019/wot/td#",
    "dct": "http://purl.org/dc/terms/",
    "properties": {
      "@id": "td:hasPropertyAffordance",
      "@type": "@id",
      "@container": "@index",
      "@context": {}
    },
    "title": "dct:title"
  }
}
```

prevents "title" from being used in the output:

```json
{
  "@context": {
    "td": "https://www.w3.org/2019/wot/td#",
    "dct": "http://purl.org/dc/terms/",
    "properties": {
      "@id": "td:hasPropertyAffordance",
      "@type": "@id",
      "@container": "@index",
      "@context": {}
    },
    "title": "dct:title"
  },
  "title": "46203",
  "properties": {
    "@none": {
      "dct:title": "Switch"
    }
  }
}
```

Instead, it uses "dct:title". Removing `"@context": {}` from the context makes it work and use "title" instead.

gkellogg:
[...]

PR #411, opened 2020-09-20
fix(#394): Clone active context to prevent mutation

RoboPhred:
Fixes issue #394 

Change was tested against the w3c test suite and seems to fix the issue.


RoboPhred:
This is being blocked by additional tests that are failing in master.  I do not know how to address these, and have opened a new issue for them: #412

PR #421, opened 2020-11-17
fix: states which protected term is being redefined

kdenhartog:
I've been fighting this error blindly for too long. It's time we shed some light on which terms are causing conflicts when protected.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -922,7 +922,7 @@ api.createTermDefinition = ({
       const protectedMode = (options && options.protectedMode) || 'error';
       if(protectedMode === 'error') {
         throw new JsonLdError(
-          'Invalid JSON-LD syntax; tried to redefine a protected term.',
+          `Invalid JSON-LD syntax; tried to redefine ${term} which is a protected term.`,
           'jsonld.SyntaxError',
           {code: 'protected term redefinition', context: localCtx, term});
       } else if(protectedMode === 'warn') {
--- a/lib/context.js
+++ b/lib/context.js
@@ -922,7 +922,7 @@ api.createTermDefinition = ({
       const protectedMode = (options && options.protectedMode) || 'error';
       if(protectedMode === 'error') {
         throw new JsonLdError(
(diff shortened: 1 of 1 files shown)


OR13 (approved):

lib/context.js
```diff
@@ -922,7 +922,7 @@ api.createTermDefinition = ({
       const protectedMode = (options && options.protectedMode) || 'error';
       if(protectedMode === 'error') {
         throw new JsonLdError(
-          'Invalid JSON-LD syntax; tried to redefine a protected term.',
+          `Invalid JSON-LD syntax; tried to redefine ${term} which is a protected term.`,
```
dlongley, line 925:
```suggestion
          `Invalid JSON-LD syntax; tried to redefine "${term}" which is a protected term.`,
```

dlongley (approved):
Approving with my minor nit comment. Thanks! cc: @davidlehn

gannan08 (commented):
Minor nit addressed: https://github.com/digitalbazaar/jsonld.js/pull/421/commits/0a8dab176e2b25fa1a16cff8e4e1cc3487f34fa7

Needs review: @davidlehn

gannan08 (approved):

davidlehn:
What is the use case for this?  I'm guessing it's something like displaying the error in a UI like the json-ld.org playground.  The term is added to the error details so it is possible to access with `err.details.term`.  I hesitate to have this one place where we construct the error message with specific details.  Here and elsewhere we've been using the details object to hold extra info depending as appropriate.  Those details are something that should be exposed in the playground if that was the issue here.

kdenhartog:
The intent is to be able to show which protected term is causing issues in tests being run in CI pipelines like the security v3 context. Typically the full objects aren't returned by jest, but it will return the full error message meaning that if the term is included it can be seen easily within the tests run.

Ideally, we should do this type of change for the variety of different errors that get thrown during context processing, but this was the only one I was focused on when trying to figure out the issues with usage of protected in security v3.

dmitrizagidulin:
Merging. Thanks again, @kdenhartog!

PR #427, opened 2021-01-21
Update rdf-canonize.

davidlehn:


davidlehn:
@dlongley `setImmediate` polyfill was removed in rdf-canonize and now karma tests fail.  Should that be restored there or should the webpack/karma setups be changed to use the builtin polyfill?

dlongley:
The webpack/karma setups should be changed to use the builtin polyfill.

davidlehn:
Ok.  Such polyfills are removed in webpack 5, so when that migration happens we'll have to use https://github.com/yuzujs/setImmediate directly.

davidlehn:
Upon further consideration, going to go with the `setimmediate` package directly in `rdf-canonize` to hopefully simplify things so packages like this don't have to do anything and will be ready for webpack v5.

Issue #420, opened 2020-11-16
Please replace request


guimard:
Hi,

[request has been deprecated](https://github.com/request/request/issues/3142), could you replace it by another library ?

Cheers,
Xavier

aljones15:
I assigned to me, I will try to replace `request` with [`ky`](https://www.npmjs.com/package/ky) ASAP.

dlongley:
@aljones15, we can use [@digitalbazaar/http-client](https://github.com/digitalbazaar/http-client).

aljones15:
Please note this issue is also related to: https://github.com/digitalbazaar/jsonld.js/issues/200

Which is to replace the documentLoader with a solution that uses the browser's `fetch` api such as axios.
In our case we will be using https://github.com/digitalbazaar/http-client which in turn uses `ky`.

We also need to fix the tests for the documentLoaders:

1. test should rely on an `express` server we start at the beginning of the test suite
2. tests should attempt to resolve contexts/ jsonld documents served as static assets from the express instance.

dmitrizagidulin:
Just to update: This is still on our radar, will get to this as time permits.

davidlehn:
Fixed in 5.0.0.

PR #434, opened 2021-02-18
Replace request issue 420

aljones15:
This replaces the deprecated library `request` with `@digitalbazaar/http-client`.
This was done a way that should ensure this is a minor feature and not a major update.
Unfortunately `node-fetch` requires node >= 10 so this is now a major release.
When we do the next major update we pull out the callbacks and such then.

Test:
The existing test suite reports no new failures, but does not use the documentLoader when testing.
I did manual tests using the compact feature:

no redirects:
```js
jsonld.compact({"http://purl.org/dc/terms/created": new Date().toISOString()}, 'https://w3c-ccg.github.io/did-spec/contexts/did-v1.jsonld').then(console.log)
Promise { <pending> }
> {
  '@context': 'https://w3c-ccg.github.io/did-spec/contexts/did-v1.jsonld',
  'dc:created': '2021-02-18T19:40:57.855Z'
}
```
302 Found redirect with a `header.location`:
```js
> jsonld.compact({"http://purl.org/dc/terms/created": new Date().toISOString()}, 'https://w3id.org/did/v1').then(console.log)
Promise { <pending> }
> {
  '@context': 'https://w3id.org/did/v1',
  'dc:created': '2021-02-18T20:00:21.644Z'
}
```
redirect with a link header containing a context:
```js
 nodeDocumentLoader('https://json-ld.org/test-suite/tests/remote-doc-0010-in.json').then(console.log);
{ contextUrl: 'remote-doc-0010-context.jsonld',
  documentUrl:
   'https://json-ld.org/test-suite/tests/remote-doc-0010-in.json',
  document: '[\n  {\n    "@id": "",\n    "term": "value"\n  }\n]' }
```

addresses:
https://github.com/digitalbazaar/jsonld.js/issues/420


package.json
```diff
@@ -118,6 +118,7 @@
   },
   "browser": {
     "lib/index.js": "./lib/jsonld.js",
+    "lib/documentLoaders/node": false,
```
davidlehn, line 121: I think its preferred to have an explicit extension:
```suggestion
    "lib/documentLoaders/node.js": false,
```

lib/documentLoaders/node.js
```diff
@@ -162,14 +143,22 @@ module.exports = ({
...
-        resolve({res, body});
-      }
-    });
-  });
+async function _fetch({url, headers, strictSSL}) {
+  try {
+    const agent = new https.Agent({rejectUnauthorized: strictSSL});
+    const res = await httpClient.get(
+      url, {headers, agent, redirect: 'manual'});
+    const body = JSON.stringify(res.data, null, 2);
```
dlongley, line 151: We should not reserialize the body here. Either we should ensure that a parsed body can be passed through or we should set flags to prevent parsing. My preference is for the former.

package.json
```diff
@@ -81,7 +81,7 @@
     "webpack-merge": "^4.2.1"
   },
   "engines": {
-    "node": ">=6"
+    "node": ">=10"
```
dlongley, line 84: I think we should actually just go to node 12 here. Node 10 expires in ~30 days or so.

dlongley (changes requested):

package.json
davidlehn: Updated to 12.

lib/documentLoaders/node.js
davidlehn: Looks like we can just avoid the stringify.  The result would just be checked if it's a string and parsed again.

PR #445, opened 2021-04-05
Remove manually defining browser global for export.

dmitrizagidulin:
Removes `global` usage (this is messing up downstream webpack and karma usage), no longer manually define the `JsonLdProcessor` global for browsers (this is left for bundlers to do).

Also fixes issue https://github.com/digitalbazaar/jsonld.js/issues/267, and the problem PR https://github.com/digitalbazaar/jsonld.js/pull/311 is trying to address.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # jsonld ChangeLog
 
+## 6.0.0 -
+
+### Changed
+- **BREAKING**: No longer manually define browser global `JsonLdProcessor`;
+  this is left to Webpack and other bundlers. (Currently, this is messing up
+  downstream bundling and karma tests.)
+
 ## 5.0.0 - 2021-03-18
 
 ### Notes
--- a/lib/jsonld.js
+++ b/lib/jsonld.js
@@ -1005,16 +1005,6 @@ jsonld.RequestQueue = require('./RequestQueue');
 /* WebIDL API */
 jsonld.JsonLdProcessor = require('./JsonLdProcessor')(jsonld);
 
-// setup browser global JsonLdProcessor
-if(_browser && typeof global.JsonLdProcessor === 'undefined') {
-  Object.defineProperty(global, 'JsonLdProcessor', {
-    writable: true,
-    enumerable: false,
-    configurable: true,
-    value: jsonld.JsonLdProcessor
-  });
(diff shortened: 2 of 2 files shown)


dmitrizagidulin:
@davidlehn - can you help resolve the webpack and browserify tests?

lib/jsonld.js
```diff
@@ -1005,16 +1003,6 @@ jsonld.RequestQueue = require('./RequestQueue');
 /* WebIDL API */
 jsonld.JsonLdProcessor = require('./JsonLdProcessor')(jsonld);
 
-// setup browser global JsonLdProcessor
-if(_browser && typeof global.JsonLdProcessor === 'undefined') {
```
dlongley, line 1009: Can this just be changed to `globalThis` to avoid a breaking change?

dmitrizagidulin: Easiest thing to do would be to change `global` to `window`. Would you prefer that?

davidlehn: `globalThis` exists for this purpose, let's use it.  That was the plan in my head, it just hadn't happened yet.  Now with the node >=12 req, and modern bundler tools, it should be easier than before.  Using `window` won't work in web workers that just have `self`.

davidlehn: Also on the TODO list was to break out node/browser code using modern practices (that detection code is ancient).  I think it's an easy change here if that makes a difference for whatever problem this code was causing.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # jsonld ChangeLog
 
+## 6.0.0 -
+
+### Changed
+- **BREAKING**: No longer manually define browser global `JsonLdProcessor`;
+  this is left to Webpack and other bundlers. (Currently, this is messing up
+  downstream bundling and karma tests.)
```
davidlehn, line 8: The suggestion below is to use `globalThis`. But if not, I don't understand what "left to Webpack and other bundlers" means.  How would that work?  Also a better explanation than "messing up downstream" would be preferred.  It's not clear what that means here or in commits.

davidlehn:
Hopefully https://github.com/digitalbazaar/jsonld.js/pull/447 addresses this for other projects that were having difficulty.

PR #447, opened 2021-04-07
Use globalThis and split out platform support.

davidlehn:
- Use `globalThis` to set globals in browser contexts.
- Split platform support out into Node.js and browser files.

Alternate approach to https://github.com/digitalbazaar/jsonld.js/pull/445.


dmitrizagidulin (approved):
:+1:

dlongley (approved):
Great, thanks!

Issue #199, opened 2017-09-12
Add feature to throw error when unmapped elements are dropped during processing


dlongley:
When certain elements in a document cannot be mapped to absolute URLs (via `@context` or a base URL), then are silently dropped during expansion per the spec. There should be a feature in this processor that, instead, causes an error to be raised. This is particularly important when this library is used in conjunction with digital signing libraries.

Related to #178.

dlongley:
Partially supported with `expansionMap` and `compactionMap` here: bcae96153a1e7da6669eeaf25f856ab93d833cd8

Still need something for `toRDF`.

dlongley:
Specifically, when converting to RDF, relative URIs are dropped and this should be handled by another map or similar mechanism.

OR13:
@dlongley I want to make this explode, whats the easiest way to do that?

I have this PR with a failing test: https://github.com/transmute-industries/vc.js/pull/26

dlongley:
@OR13,

>  I want to make this explode, whats the easiest way to do that?

Sorry, just seeing this now, a month later. Using an undefined `type` with no base set or -- as it says in the issue I just linked to -- a non-URI for an `id` will cause silent droppage. Not sure if that's "exploding" though. :)

OR13:
explode means that when someone hands json that has been mutated after being signed, a verifier will throw an error on it, not return "verified".

PR #452, opened 2021-05-17
Expand expansion map to support calling when a relative IRI is detected

tplooker:
As documented in #199 when the value of `@type` or `@id` is a relative IRI it is silently dropped when the resulting expanded object is converted to an RDF representation because RDF does not support relative IRIs. This is problematic for cases where a JSON-LD document is being digitally signed as information can be silently dropped.

This PR expands the `expansionMap` function so that it is called when a relative IRI is encountered when expanding an `@id` or `@type` value. This then allows libraries like [jsonld-signatures](https://github.com/digitalbazaar/jsonld-signatures) to throw errors through a custom [strict expansion map](https://github.com/digitalbazaar/jsonld-signatures/blob/master/lib/expansionMap.js#L6-L13).

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -1043,11 +1043,11 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
 
   // prepend vocab
   if(relativeTo.vocab && '@vocab' in activeCtx) {
-    return activeCtx['@vocab'] + value;
+    value = activeCtx['@vocab'] + value;
   }
 
   // prepend base
-  if(relativeTo.base && '@base' in activeCtx) {
+  else if(relativeTo.base && '@base' in activeCtx) {
     if(activeCtx['@base']) {
       // The null case preserves value as potentially relative
       value = prependBase(prependBase(options.base, activeCtx['@base']), value);
@@ -1061,6 +1061,7 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
     const expandedResult = options.expansionMap({
       relativeIri: value,
       activeCtx,
+      options
     });
     if(expandedResult !== undefined) {
       value = expandedResult;
--- a/lib/context.js
+++ b/lib/context.js
@@ -1041,6 +1041,14 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
     }
   }
 
+  // A flag that captures whether the iri being expanded is
+  // the value for an @type
+  let typeExpansion = false;
+
+  if (options !== undefined && options.typeExpansion !== undefined) {
+    typeExpansion = options.typeExpansion;
+  }
+
   if(relativeTo.vocab && '@vocab' in activeCtx) {
     // prepend vocab
     const prependedResult = activeCtx['@vocab'] + value;
@@ -1056,7 +1064,8 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
           type: '@vocab',
           vocab: activeCtx['@vocab'],
           value,
-          result: prependedResult
+          result: prependedResult,
+          typeExpansion,
         },
         activeCtx,
         options
@@ -1078,6 +1087,9 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
       if(activeCtx['@base']) {
         base = prependBase(options.base, activeCtx['@base']);
         prependedResult = prependBase(base, value);
+      } else {
+        base = activeCtx['@base'];
+        prependedResult = value;
       }
     } else {
       base = options.base;
@@ -1094,7 +1106,8 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
           type: '@base',
           base,
           value,
-          result: prependedResult
+          result: prependedResult,
+          typeExpansion,
         },
         activeCtx,
         options
@@ -1102,7 +1115,7 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
     }
     if(expansionMapResult !== undefined) {
       value = expansionMapResult;
-    } else if(prependedResult !== undefined) {
+    } else {
       // the null case preserves value as potentially relative
       value = prependedResult;
     }
@@ -1118,6 +1131,7 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
     const expandedResult = options.expansionMap({
       relativeIri: value,
       activeCtx,
+      typeExpansion,
       options
     });
     if(expandedResult !== undefined) {
--- a/lib/context.js
+++ b/lib/context.js
@@ -1041,13 +1041,11 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
     }
   }
 
-  // prepend vocab
   if(relativeTo.vocab && '@vocab' in activeCtx) {
+    // prepend vocab
     value = activeCtx['@vocab'] + value;
-  }
-
-  // prepend base
-  else if(relativeTo.base && '@base' in activeCtx) {
+  } else if(relativeTo.base && '@base' in activeCtx) {
+    // prepend base
     if(activeCtx['@base']) {
       // The null case preserves value as potentially relative
       value = prependBase(prependBase(options.base, activeCtx['@base']), value);
--- a/lib/context.js
+++ b/lib/context.js
@@ -1050,7 +1050,7 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
   if(relativeTo.base && '@base' in activeCtx) {
     if(activeCtx['@base']) {
       // The null case preserves value as potentially relative
-      return prependBase(prependBase(options.base, activeCtx['@base']), value);
+      value = prependBase(prependBase(options.base, activeCtx['@base']), value);
     }
   } else if(relativeTo.base) {
     value = prependBase(options.base, value);
--- a/lib/context.js
+++ b/lib/context.js
@@ -1045,7 +1045,7 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
   // the value for an @type
   let typeExpansion = false;
 
-  if (options !== undefined && options.typeExpansion !== undefined) {
+  if(options !== undefined && options.typeExpansion !== undefined) {
     typeExpansion = options.typeExpansion;
   }
 
--- a/lib/context.js
+++ b/lib/context.js
@@ -1043,18 +1043,96 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
 
   if(relativeTo.vocab && '@vocab' in activeCtx) {
     // prepend vocab
-    value = activeCtx['@vocab'] + value;
+    const prependedResult = activeCtx['@vocab'] + value;
+    let expansionMapResult = undefined;
+    if(options && options.expansionMap) {
+      // if we are about to expand the value by prepending
+      // @vocab then call the expansion map to inform
+      // interested callers that this is occurring
+
+      // TODO: use `await` to support async
+      expansionMapResult = options.expansionMap({
+        prependedIri: {
+          type: '@vocab',
+          vocab: activeCtx['@vocab'],
+          value,
+          result: prependedResult
+        },
+        activeCtx,
+        options
+      });
+
+    }
+    if(expansionMapResult !== undefined) {
+      value = expansionMapResult;
+    } else {
+      // the null case preserves value as potentially relative
+      value = prependedResult;
+    }
   } else if(relativeTo.base && '@base' in activeCtx) {
     // prepend base
     if(activeCtx['@base']) {
-      // The null case preserves value as potentially relative
-      value = prependBase(prependBase(options.base, activeCtx['@base']), value);
+      const prependedResult = prependBase(
+        prependBase(options.base, activeCtx['@base']), value);
+
+      let expansionMapResult = undefined;
+      if(options && options.expansionMap) {
+        // if we are about to expand the value by prepending
+        // @base then call the expansion map to inform
+        // interested callers that this is occurring
+
+        // TODO: use `await` to support async
+        expansionMapResult = options.expansionMap({
+          prependedIri: {
+            type: '@base',
+            base: activeCtx['@base'],
+            value,
+            result: prependedResult
+          },
+          activeCtx,
+          options
+        });
+      }
+      if(expansionMapResult !== undefined) {
+        value = expansionMapResult;
+      } else {
+        // the null case preserves value as potentially relative
+        value = prependedResult;
+      }
     }
   } else if(relativeTo.base) {
-    value = prependBase(options.base, value);
+    const prependedResult = prependBase(options.base, value);
+    let expansionMapResult = undefined;
+    if(options && options.expansionMap) {
+      // if we are about to expand the value by prepending
+      // @base then call the expansion map to inform
+      // interested callers that this is occurring
+
+      // TODO: use `await` to support async
+      expansionMapResult = options.expansionMap({
+        prependedIri: {
+          type: '@base',
+          base: options.base,
+          value,
+          result: prependedResult
+        },
+        activeCtx,
+        options
+      });
+    }
+    if(expansionMapResult !== undefined) {
+      value = expansionMapResult;
+    } else {
+      value = prependedResult;
+    }
   }
 
-  if(!_isAbsoluteIri(value) && options.expansionMap) {
+  if(!_isAbsoluteIri(value) && options && options.expansionMap) {
+    // if the result of the expansion is not an absolute iri then
+    // call the expansion map to inform interested callers that
+    // the resulting value is a relative iri, which can result in
+    // it being dropped when converting to other RDF representations
+
     // TODO: use `await` to support async
     const expandedResult = options.expansionMap({
       relativeIri: value,
--- a/lib/context.js
+++ b/lib/context.js
@@ -1069,42 +1069,22 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
       // the null case preserves value as potentially relative
       value = prependedResult;
     }
-  } else if(relativeTo.base && '@base' in activeCtx) {
-    // prepend base
-    if(activeCtx['@base']) {
-      const prependedResult = prependBase(
-        prependBase(options.base, activeCtx['@base']), value);
-
-      let expansionMapResult = undefined;
-      if(options && options.expansionMap) {
-        // if we are about to expand the value by prepending
-        // @base then call the expansion map to inform
-        // interested callers that this is occurring
-
-        // TODO: use `await` to support async
-        expansionMapResult = options.expansionMap({
-          prependedIri: {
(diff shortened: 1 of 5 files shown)


dlongley:
@tplooker, Thanks heaps! A PR to address this has been a long time coming.

The implementation here has me a bit worried that we may be missing some other cases though, based on it explicitly targeting `@id` and `@type` in `_expandObject`. For example, I think it will miss the usage of `@type` in values. Perhaps it should instead target `_expandIri`. So we'd always pass the `expansionMap` in there (perhaps as another processing option), and then use the new code from this PR when `base: true` is in `relativeTo` (and we may also need to cover when vocab is relative and falls back to using base).

@davidlehn, what are your thoughts?

OR13:
Super excited to start throwing errors instead of silently dropping terms... this is huge.

tplooker:
@dlongley @davidlehn I have updated the implementation as per your suggestion @dlongley, can you suggest some further test cases to verify this behaviour?

tplooker:
Outstanding questions
- Do we need similar protections around `@vocab` usage?
- Do we need to bubble up more info other than the value of the relative Iri in the expansion map, such as the property (if there is one) it pertains to? So that a strict expansion map can suitably error report when the relative iri is for an `@type` and `@id` property

lib/expand.js
```diff
@@ -420,6 +420,9 @@ async function _expandObject({
   const nests = [];
   let unexpandedValue;
 
+  // Add expansion map to the processing options
+  options = {...options, expansionMap};
```
ivy-wei-yang, line 424: Do we need to pass `expansionMap` from `expand` method to `_expandValue` options, which is then passed to `_expandIri` in downstream?

dlongley: Yes, `_expandIri` should always receive the `expansionMap` if it was passed in -- otherwise we're going to miss corner cases.

dlongley:
> Do we need similar protections around @vocab usage?

Since `@vocab` can be relative and depend on `@base` (via `@context` or via API options), we'll need to ensure we cover that case (I didn't look at the code to see what the simplest change is for that case). We'll want a test for this case. Ultimately, we'll want `expansionMap` tests for relative URIs that appear with any keyword (whether in a node object or literal value) that accepts them.

> Do we need to bubble up more info other than the value of the relative Iri in the expansion map, such as the property (if there is one) it pertains to? So that a strict expansion map can suitably error report when the relative iri is for an `@type` and `@id` property?

That does sound useful.

tplooker:
> Since @vocab can be relative and depend on @base (via @context or via API options), we'll need to ensure we cover that case (I didn't look at the code to see what the simplest change is for that case). We'll want a test for this case. Ultimately, we'll want expansionMap tests for relative URIs that appear with any keyword (whether in a node object or literal value) that accepts them.

Ok I have added a test for when `@vocab` has a value of `./` but im not sure I have got all other cases yet.

> That does sound useful.

Ok any suggestion on the approach here because at the moment the term/property the value is being expanded against in the expandIri function is not in scope so we would need to thread this in from every possible call of the expandIri function?

dlongley:
> Ok any suggestion on the approach here because at the moment the term/property the value is being expanded against in the expandIri function is not in scope so we would need to thread this in from every possible call of the expandIri function?

Yeah, let's punt that to a future PR if it becomes clear that users really need the additional information (we're just guessing right now). It's more important to enable some kind of error to be thrown here via expansion maps.

lib/expand.js
```diff
@@ -75,6 +75,10 @@ api.expand = async ({
   typeScopedContext = null,
   expansionMap = () => undefined
 }) => {
+
+  // Add expansion map to the processing options
```
dlongley, line 79:
```suggestion
  // add expansion map to the processing options
```

lib/context.js
```diff
@@ -1043,17 +1043,29 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
 
   // prepend vocab
   if(relativeTo.vocab && '@vocab' in activeCtx) {
-    return activeCtx['@vocab'] + value;
+    value = activeCtx['@vocab'] + value;
   }
 
   // prepend base
-  if(relativeTo.base && '@base' in activeCtx) {
+  else if(relativeTo.base && '@base' in activeCtx) {
```
dlongley, line 1050: This won't let me suggest here, but the style needs to be cleaned up to be something like this:

```js
    value = activeCtx['@vocab'] + value;
  } else if(relativeTo.base && '@base' in activeCtx) {
    // prepend base
```

lib/expand.js
```diff
@@ -420,6 +424,9 @@ async function _expandObject({
   const nests = [];
   let unexpandedValue;
 
+  // Add expansion map to the processing options
```
dlongley, line 427:
```suggestion
  // add expansion map to the processing options
```

tests/misc.js
```diff
@@ -478,3 +478,266 @@ describe('literal JSON', () => {
     });
   });
 });
+
+describe('expansionMap', () => {
+  it('should be called on un-mapped term', async () => {
```
dlongley, line 483:
```suggestion
  it('should be called on unmapped term', async () => {
```

tests/misc.js
```diff
@@ -478,3 +478,266 @@ describe('literal JSON', () => {
...
+        expansionMapCalled = true;
+      }
+    };
+
+    await jsonld.expand(docWithUnMappedTerm, {expansionMap});
+
+    assert.equal(expansionMapCalled, true);
+  });
+
+  it('should be called on nested un-mapped term', async () => {
```
dlongley, line 504:
```suggestion
  it('should be called on nested unmapped term', async () => {
```

tests/misc.js
```diff
@@ -478,3 +478,266 @@ describe('literal JSON', () => {
...
+    const expansionMap = info => {
+      if(info.relativeIri === 'relativeiri') {
+        expansionMapCalled = true;
+      }
+    };
+
+    await jsonld.expand(docWithRelativeIriId, {expansionMap});
+
+    assert.equal(expansionMapCalled, true);
+  });
```
dlongley, line 612: Trying to add (what I believe) is a missing test on typed literals via suggest mode here after this other test:

```suggestion
  it('should be called on relative iri for type term', async () => {
    const docWithRelativeIriId = {
      '@context': {
        'definedTerm': 'https://example.com#definedTerm'
      },
      'id': "urn:absoluteiri",
      '@type': "relativeiri",
      definedTerm: "is defined"
    };

    let expansionMapCalled = false;
    const expansionMap = info => {
      if(info.relativeIri === 'relativeiri') {
        expansionMapCalled = true;
      }
    };

    await jsonld.expand(docWithRelativeIriId, {expansionMap});

    assert.equal(expansionMapCalled, true);
  });
  
  it('should be called on relative iri for typed literal', async () => {
    const docWithRelativeIriId = {
      '@context': {
        'definedTerm': 'https://example.com#definedTerm'
      },
      'id': "urn:absoluteiri",
      definedTerm: {"@type": "relativeiri", "@value": "is defined"}
    };

    let expansionMapCalled = false;
    const expansionMap = info => {
      if(info.relativeIri === 'relativeiri') {
        expansionMapCalled = true;
      }
    };

    await jsonld.expand(docWithRelativeIriId, {expansionMap});

    assert.equal(expansionMapCalled, true);
  });  
```

dlongley (approved):
Thanks, @tplooker!

Approving with some style-nits and a suggested test for typed literals.

@davidlehn -- can you please take a look at this?

tests/misc.js
tplooker: Interesting @dlongley when I run the test you added locally I get `jsonld.SyntaxError: Invalid JSON-LD syntax; an element containing "@value" and "@type" must have an absolute IRI for the value of "@type".` is there a clear delineation between when the expanded value of `@type` results in a relativeIri  is considered a syntax violation vs it not being?

lib/context.js
tplooker: Updated locally

tests/misc.js
dlongley: @tplooker,

I think the "clear delineation" is probably exactly the error message you quoted. You can't use a relative `@type` value in a typed literal value (vs. a node object). What we want to be testing there is what happens if a base URL is also passed in:

```js
await jsonld.expand(docWithRelativeIriId, {expansionMap, base: 'https://example.com'});
```

Hmm, this is also pointing to the fact that the current implementation in this PR isn't fully going to do what we want for signatures. We want the `expansionMap` to be called before a base is prepended -- and probably include the `base` value in the options passed to the handler function so an implementation can decide if it cares at that point.

We would certainly want jsonld-signatures lib's `expansionMap` to throw happen when expanding `@type` values (regardless of `base`'s value) -- because it is a common case that when people forget to define those in a document they are signing, they don't want any base to be applied to them giving them an unexpected (and wrong) definition.

dlongley: Thinking about this more...

While it would be easy enough to call the `expansionMap` prior to prepending base -- it will be tricky to get the property associated with the relative IRI value passed to the handler without a lot of changes (i.e., every call of `expandIri`) as we discussed above.

We could proceed with what we have for now -- and only throw when base isn't set -- which will cover *most* use cases. We recommend not setting `@base` in the JSON-LD signature world ... but now that I think about it, I recall @OR13 trying to use it somewhere. If someone uses `@base` *and* forgets to define an `@type` value, they could get burned because the URL will resolve to an absolute one (using their `@base` value) that may not be the one they intended.

So -- our options here are to make it so the `expansionMap` handler function gets the information it needs to decide. For example, we may want the logic in the jsonld-signatures `expansionMap` handler to be:

Is a relative `@type` value being expanded (prior to prepending base)? If so, throw an error. If not, see if a `base` is set. If one is set, don't throw, otherwise throw.

How do you think we should proceed?

dlongley: We could do something (maybe?) hacky and just pass whether an `@type` value is being expanded (a boolean value) to `expandIri` and default it to false. Then we'd only need to pass this where `@type` expansion occurs vs. having to update `expandIri` everywhere. That would cover the common use case -- and the only use case. All other URI values should be unrelated to vocabularies -- they'd just be purely node identifier values.

This approach may be the simplest way to fix it and cover most (all?) cases.

dlongley: Yeah, it would be good to expose whenever `@base` or `@vocab` are going to be applied. And I think my suggestion of saying whether or not it's an `@type` expansion should probably be sufficient (vs. having to pass the associated property for every value).

tplooker:
@dlongley @davidlehn I have now added another expansionMap hook called "prependIri" which is fired whenever a values expansion is going to occur that will involve it being prepend with either the value of '@base' or '@vocab'. This should allows us at the JSON-LD signatures layer to build an expansion map that at the very least warns a caller that an expansion occured using `@base` which __could__ be un-expected in certain cases, such as the case outlined [here](https://github.com/digitalbazaar/jsonld.js/pull/452#discussion_r636397247)

tests/misc.js
tplooker: See [here](https://github.com/digitalbazaar/jsonld.js/pull/452#issuecomment-845514850) for an update

lib/context.js
```diff
@@ -1041,19 +1041,107 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
...
+    const prependedResult = prependBase(options.base, value);
+    let expansionMapResult = undefined;
+    if(options && options.expansionMap) {
+      // if we are about to expand the value by prepending
+      // @base then call the expansion map to inform
+      // interested callers that this is occurring
+
+      // TODO: use `await` to support async
+      expansionMapResult = options.expansionMap({
+        prependedIri: {
```
ivy-wei-yang, line 1113: Is there any doc on expansionMap that what data will be passed to the hook? Do users have to read the code or unit tests to understand?

lib/expand.js
```diff
@@ -420,6 +424,9 @@ async function _expandObject({
   const nests = [];
   let unexpandedValue;
 
+  // add expansion map to the processing options
+  options = {...options, expansionMap};
```
ivy-wei-yang, line 428: Since expansionMap is passed to `options` from top level `api.expand`, here might be redundant

lib/expand.js
```diff
@@ -75,6 +75,10 @@ api.expand = async ({
   typeScopedContext = null,
   expansionMap = () => undefined
 }) => {
+
+  // add expansion map to the processing options
+  options = {...options, expansionMap};
+
```
ivy-wei-yang, line 81: 👍

lib/context.js
dlongley: Nothing at the moment, but PRs welcome to add something to the README.

lib/context.js
```diff
@@ -1041,19 +1041,107 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
...
+    }
+    if(expansionMapResult !== undefined) {
+      value = expansionMapResult;
+    } else {
+      // the null case preserves value as potentially relative
+      value = prependedResult;
+    }
+  } else if(relativeTo.base && '@base' in activeCtx) {
+    // prepend base
     if(activeCtx['@base']) {
```
dlongley, line 1074: I think we can simplify the logic here and consolidate the new code into one place with a single conditional on just `relativeTo.base` at the top-level ... and then getting the base value from `@base in activeCtx` and falling back to `options.base`. That would also make sure we still use the `expansionMap` when `activeCtx['@base']` is `null` ... because I think we may be failing to do so here at the moment.

@davidlehn -- any feedback on what we're passing the `expansionMap` handler here?

dlongley (changes requested):

dlongley:
@tplooker -- I think the approach here is ok and I've asked for bikeshedding feedback (if any) on the naming. See my comment ... I noticed we may have a gap when base is `null` that we could avoid by also making the code more DRY.

lib/expand.js
tplooker: Reviewing this again, I agree, currently the only calls to _expandObject are coming from expand however if there are other callers of this function in future it might not hurt to have this extra layer of protection.

lib/context.js
tplooker: @dlongley thanks I agree, have updated the PR accordingly.

tplooker:
> @tplooker -- I think the approach here is ok and I've asked for bikeshedding feedback (if any) on the naming. See my comment ... I noticed we may have a gap when base is null that we could avoid by also making the code more DRY.

Ready for review again, can you elaborate on this case some more? We have coverage over the case when `activeCtx['@base']` is null but not when `options.base` is, is that what you mean?

lib/context.js
```diff
@@ -1041,19 +1041,88 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
...
-    return prependBase(options.base, value);
+    // prepend base
+    let prependedResult;
+    let expansionMapResult;
+    let base;
+    if('@base' in activeCtx) {
+      if(activeCtx['@base']) {
+        base = prependBase(options.base, activeCtx['@base']);
+        prependedResult = prependBase(base, value);
+      }
```
dlongley, line 1081:
```suggestion
      if(activeCtx['@base']) {
        base = prependBase(options.base, activeCtx['@base']);
        prependedResult = prependBase(base, value);
      } else {
        // falsy (`''` or `null`) `@base` preserves value as potentially relative
        base = activeCtx['@base'];
        prependedResult = value;
      }
```

lib/context.js
```diff
@@ -1041,19 +1041,88 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
...
+          result: prependedResult
+        },
+        activeCtx,
+        options
+      });
+    }
+    if(expansionMapResult !== undefined) {
+      value = expansionMapResult;
+    } else if(prependedResult !== undefined) {
+      // the null case preserves value as potentially relative
```
dlongley, line 1106:
```suggestion
    } else {
```

dlongley (changes requested):
@tplooker,

> Ready for review again, can you elaborate on this case some more? We have coverage over the case when activeCtx['@base'] is null but not when options.base is, is that what you mean?

What I mean is that we don't call the `expansionMap` prepend handler when `activeCtx['@base']` is falsy -- which means relative IRIs can sneak in without the handler having the opportunity to evaluate whether the value was to be expanded using `@vocab` or `@base` or whether the value was associated with an `@type`.

Additionally, unless I missed it, I didn't see where we signal whether the value is associated with `@type` or not. This will be key information for use cases such as jsonld-signatures in determining whether or not to throw an error -- as I fully expect us to want to throw if `@base` is being applied to a value associated with `@type`. We should add a flag such as `typeExpansion` that can indicate whether the value is associated with `@type`, default it to `false` and only set it to true where we are expanding `@type` values.

tplooker:
@dlongley thanks I think I've addressed all of the points you raised in your last review.

lib/context.js
```diff
@@ -1041,19 +1041,102 @@ function _expandIri(activeCtx, value, relativeTo, localCtx, defined, options) {
...
   }
 
-  // prepend vocab
-  if(relativeTo.vocab && '@vocab' in activeCtx) {
-    return activeCtx['@vocab'] + value;
+  // A flag that captures whether the iri being expanded is
+  // the value for an @type
+  let typeExpansion = false;
+
+  if (options !== undefined && options.typeExpansion !== undefined) {
```
dlongley, line 1048: Style nit:

```suggestion
  if(options !== undefined && options.typeExpansion !== undefined) {
```

dlongley (approved):
Thanks, @tplooker, this is awesome! LGTM -- @davidlehn any thoughts before this gets pulled in to finally address #199?

tplooker:
Hey @dlongley @davidlehn anything outstanding with this PR, keen to merge when possible so I can continue to work upstream?

dlongley:
@tplooker -- Just letting you know that I see this and that I'm going to get @davidlehn to have a look.

davidlehn:
- I have been following along, but wasn't sure how to review.  I think we'll put this out there and iterate later if needed.
- The lack of docs on these sorts of features is a problem, but that's a larger issue to tackle later.  We'll at least have the tests for people to look at.
- I'll add the changelog entry.  In  the future, nice to add that in the PRs.
- There's probably some overlap with what I was trying to do in https://github.com/digitalbazaar/jsonld.js/pull/366.  Will have to explore if these features can be combined somehow later.

davidlehn (approved):

tests/misc.js
```diff
@@ -478,3 +478,531 @@ describe('literal JSON', () => {
...
+      assert.equal(expansionMapCalled, true);
+    });
+  });
+
+  describe('prependedIri', () => {
+    it("should be called when property is \
+    being expanded with `@vocab`", async () => {
+      const doc = {
+        '@context': {
+          "@vocab": "http://example.com/",
```
OR13, line 819: was there a release that included this?

dlongley: @OR13 -- a release was made with the new `safe` mode that does a more low-level and targeted approach to solving the "undefined term / relative URL" problem with a much simpler API (set `safe: true` in the options). The latest version of jsonld.js (v8) now sets `safe: true` by default in `canonize` so that method will fail closed if you don't define your terms or use absolute URLs. You can override the default settings of course, as usual.

PR #460, opened 2021-08-13
Improve compactIri performance

Janpot:
Having some performance problems on some particularly large input. I noticed this quite costly loop that could be replaced with a Map lookup. Seems to reduce the overall runtime by about 50%.

for reference, the offending data: https://gist.githubusercontent.com/Janpot/792ae65f50d53a2903ffd87d29bec782/raw/51199036e09c9567ebfa4b04d03f5e207a39cdae/large-jsonld.json
encountered on this website: https://www.stanglwirt.com

### DIFF
--- a/lib/compact.js
+++ b/lib/compact.js
@@ -929,13 +929,13 @@ api.compactIri = ({
 
   // If iri could be confused with a compact IRI using a term in this context,
   // signal an error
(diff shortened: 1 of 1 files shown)


dlongley (approved):
@davidlehn, this seems sensible to me -- can we get it merged and closed?

PR #462, opened 2021-09-16
Add support for xsd:float as native type

pduchesne:
Currently, when parsing RDF using 'useNativeTypes', xsd:float is not recognized.
This fixes it.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -27,6 +27,7 @@ module.exports = {
   XSD,
   XSD_BOOLEAN: XSD + 'boolean',
   XSD_DOUBLE: XSD + 'double',
+  XSD_FLOAT: XSD + 'float',
   XSD_INTEGER: XSD + 'integer',
   XSD_STRING: XSD + 'string',
 };
--- a/lib/fromRdf.js
+++ b/lib/fromRdf.js
@@ -326,7 +326,9 @@ function _RDFToObject(o, useNativeTypes, rdfDirection) {
         }
       }
       // do not add native type
-      if(![XSD_BOOLEAN, XSD_INTEGER, XSD_FLOAT, XSD_DOUBLE, XSD_STRING].includes(type)) {
+      // Remark: xsd:float are parsed into native type (above), but yet the type is explicitly added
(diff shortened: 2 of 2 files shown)


davidlehn:
I believe the omission of xsd:float was done on purpose.  I think it was due to how native JSON numbers are handled and problems with round trips of the data.  Or maybe some other edge cases.  It was simpler to just handle all data as doubles and map to how JSON is normally processed.  I need some help remembering the details of all this.

This kind of change would need a handful of tests.  Need to text various combinations of doubles and floats, context datatypes, serializing and deserializing the data, overflow edge cases, etc.  It's probably also something to bring up with the JSON-LD community to see if it's an issue that should have broad support across implementations.  In that case the main JSON-LD test suite and spec would need updating too.

gkellogg:
Yes, the JSON number format doesn't distinguish between Integers Double or Floats. We have some logic to distinguish between Integers and Doubles, but there is really no way to distinguish a Double from a Float.

Of course, a native number may be used in a value object where `@type` is set to `xsd:float`, for example, but only as part of a value object. This is defined by the spec, not the implementation.

pduchesne:
Thank you for your feedback.

First some context to explain my usecase : i'm using fromRDF, chained with jsonld framing, to turn third-party RDF into plain JSON, and I need native numbers in that JSON. And it'd be a shame to have to do yet another stage of postprocessing to do that number parsing.

But Indeed, I hear your point: silently converting xsd:float to native JSON numbers would lose information as it would conflate xsd:float with xsd:double.

As far as I'm concerned, I've modified my branch to still parse to native type, but keep the 'xsd:float' type. With the appropriate frame, the jsonld framing algo can then easily reduce a `{"@value": 0.01, @type: "xsd:float"}` object to a plain number.
I let you be the judges of whether that's an appropriate PR.

ioggstream:
@pduchesne I have a similar use case... not sure whether the most "interoperable" way is to pre-process the RDF and replace xsd:xxx to xsd:float

See https://github.com/digitalbazaar/pyld/issues/231

PR #464, opened 2021-12-15
Update expandMap and tests.

davidlehn:
This is a continuation of the expansionMap work from:
https://github.com/digitalbazaar/jsonld.js/pull/452

@tplooker This likely needs your input.

That expansionMap work was merged but never released. I wanted to get it out there and active for use in VCs and elsewhere to at least try and check for dropped types and such. It turns out I don't know how to use this feature! `expansionMap` is now called lots more than before. It has many variations, and the lack of documentation is now rather painful, since it's not obvious what the `expansionMap` code should do for what situations.

In the case of VCs, it seems like type scoped properties like `credentialSubject` end up with a couple calls, but maybe it shouldn't? Perhaps the callback is getting called while the algorithm does normal processing rather than just when there's an issue?

In any case, I updated the tests to count all the calls to `expansionMap`, and (maybe) what they were for.  Just so it's clear in examples what is happening.

I do think some more test cases are needed, if for nothing else than as an example. Have various inputs that have unknown properties, types, etc, and an example `expansionMap` handler that knows how to reject all that. Basically a common use case for strict signing software.

### DIFF
--- a/tests/misc.js
+++ b/tests/misc.js
@@ -480,6 +480,48 @@ describe('literal JSON', () => {
 });
 
 describe.only('expansionMap', () => {
+  // track all the counts
+  // use simple count object (don't use tricky test keys!)
+  function addCounts(counts, info) {
+    // overall call count
+    counts.expansionMap = counts.expansionMap || 0;
+    counts.expansionMap++;
+
+    if(info.unmappedProperty) {
+      const c = counts.unmappedProperty = counts.unmappedProperty || {};
+      const k = info.unmappedProperty;
+      c[k] = c[k] || 0;
+      c[k]++;
+    }
+
+    if(info.unmappedValue) {
+      const c = counts.unmappedValue = counts.unmappedValue || {};
+      const v = info.unmappedValue;
+      let k;
+      if(Object.keys(v).length === 1 && '@id' in v) {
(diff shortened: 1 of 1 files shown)


tplooker:
Hey @davidlehn, sorry totally missed your ping here, this looks like a great improvement over measuring the behaviour of the function in the tests. From memory I do remember noting that the function was called multiple times during expansion, I also remember tracing the various cases and them making sense, however unfortunately I did not write it down.

davidlehn:
@tplooker No problem.  I'm dug into the issue now, so will come up with something to address what this was aiming at.  I'm not quite sure what to do with the code as is.  It seems awkward to write an `expansionMap`.  And it's named poorly if it's being used to just handle malformed input.  But at least the call sites are there now!  It's likely some other code will be added in the same spots to handle the common error handler case.

I think I found part of the duplicated call issue.  `_expandIri` is called multiple times needlessly at one point.  I'll address that soon.

If anyone knows a use case for `expansionMap` or `compactionMap` that involves... mapping, let me know. :-)  All I know to use these hooks for is for warning and error handling.

davidlehn:
This changed and grew quite a bit in another branch.  Merging back to main through this PR.

PR #481, opened 2022-04-06
Add events and safe mode.

davidlehn:
(This is a combination of "expanionMap" work and "events" work.)
- Experimental non-standard `protectedMode` option removed.
- Event handler system added.
  - Used for features to catch specific processing conditions.
  - Useful for debugging tools.
- Events emitted for spec warning cases, dropped data, and other conditions.
- "safe" mode added that uses specific events to stop processing. Very important for tooling that needs to protect against unintended data loss.


davidlehn:
- The `compactionMap` and `expansionMap` hooks were removed in favor of `safe` mode.
- The event system behind the scenes is not being exposed yet.

CHANGELOG.md
```diff
@@ -1,14 +1,18 @@
...
 - Support benchmark output in EARL output.
 - Benchmark comparison tool.
-
-### Changed
-- Change EARL Assertor to Digital Bazaar, Inc.
-- Update eslint dependencies.
+- Add "safe mode" to all APIs. Enable by adding `{safe: true}` to API options.
+  This mode causes processing to fail when data constructs are encountered that
+  result in lossy behavior or other data warnings. This is intended to be the
+  common way that digital signing and similar applications use this libraray.
```
dlongley, line 15:
```suggestion
  common way that digital signing and similar applications use this library.
```

CHANGELOG.md
```diff
@@ -32,6 +36,13 @@
     and provide other improvements.
   - Use global `URL` interface to handle relative redirects.
 
+### Removed
+- Experimental non-standard `protectedMode` option.
+- **BREAKING**: Various console warnings were removed. The newly added "safe
+  mode" can stop processing where these warnings were.
+- **BREAKING**: Remove `compactionMap` and `expansionMap`. Their known use
+  cases are addressed with "safe mode" and future planned features.
```
dlongley, line 44: Is this section in the right place? If so, ignore. If not... move :)

lib/events.js
```diff
@@ -0,0 +1,178 @@
...
+]);
+
+// safe handler that rejects unsafe warning conditions
+api.safeEventHandler = function safeEventHandler({event, next}) {
+  // fail on all unsafe warnings
+  if(event.level === 'warning' && _notSafeEventCodes.has(event.code)) {
+    throw new JsonLdError(
+      'Safe mode validation error.',
+      'jsonld.ValidationError',
+      {event}
```
dlongley, line 136: I'm not sure what this will look like in practice, but we will want to make sure we very clearly surface the error information so it's as easy as possible for devs to see and correct. I expect this to be a no-op for now but may need additional improvement in the future.

lib/events.js
```diff
@@ -0,0 +1,178 @@
...
+      throw new JsonLdError(
+        'Invalid event handler.',
+        'jsonld.InvalidEventHandler',
+        {event});
+    }
+  }
+  return doNext;
+}
+
+const _notSafeEventCodes = new Set([
```
dlongley, line 103: I like how this is implemented. Really clean and just reuses the existing spec. Perfect for any other implementer to refer to warnings raised in the spec and bail out when they occur during a canonicalization process.

dlongley (approved):
Awesome! Thanks! Check out my comments and let's get this merged and released.

lib/events.js
davidlehn: This does need to be all written up.  I went through a few iterations of those names to get something that seemed coherent and aligned with current error code naming.  Could use wider feedback.

lib/events.js
davidlehn: Understood.  I wasn't sure what to do here so just threw in the event info direct.  This PR doesn't even expose the event system so that could be confusing.  Might make more sense to expose codes or other info directly.  Not sure.

A related work-in-progress is optional path tracking so when you do get an error it's easy to find where it was.

I think we can just leave this as is for now.

CHANGELOG.md
davidlehn: Rebase bug.  Fixed.

PR #484, opened 2022-05-03
Update to `@digitalbazaar/http-client@3`.

davidlehn:
- Pulls in newer `ky` and `ky-universal` that should address security
  alerts and provide other improvements.
- Newer `ky` will throw errors on redirects even when in `manual`
  redirect mode. Now using `throwHttpErrors` option to turn off errors.
- One of the updated dependencies can cause tests to rewrite redirect
  `Location` URLs to be relative references. The global `URL` interface
  is now used to rebuild a full URL for further redirect processing.
- Newer `node-forge` will output a one-time warning if code even
  accesses `response.data`. `@digitalbazar/http-client` will forcefully
  set `data` *if* it detects a JSON content type. Calling code can't
  know if that happened, so currently needs to redo content detection to
  know if JSON `data` can be accessed. This is an issue for sites where
  JSON-LD was requested but the response is non-JSON with a `Link`
  header pointing to the JSON-LD.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,5 +6,11 @@ module.exports = {
     es2020: true
   },
   extends: 'eslint-config-digitalbazaar',
-  root: true
+  root: true,
+  ignorePatterns: [
+    'dist/',
+    'tests/webidl/WebIDLParser.js',
+    'tests/webidl/idlharness.js',
+    'tests/webidl/testharness.js'
+  ]
 };
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [12.x, 14.x, 16.x]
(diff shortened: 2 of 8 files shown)


davidlehn:
Looks like ky-universal index.js doesn't make it past node 12 parser.  If we have to remove node 12 support, this will need to be a major release.

dlongley (approved):
The node 12 error could potentially be avoided by detecting node 12 and statically loading the web streams polyfill version `=3.0.3` (for compatibility with other libs) prior to loading ky (ky-universal is failing node 12 because it uses a top-level await to try to load that polyfill -- which is gated on `globalThis.ReadableStream`).

We should make the next major release of jsonld.js require node 14.x though.

davidlehn:
The top-level await issue was a blocker.  Node.js 12 won't even parse that syntax so pre loading libraries would not work. 
 Removing Node.js 12 support and planning a major release.  Will retarget to a v6.x branch.  The only part that doesn't work with older versions is the included node document loader.  If a solution presents itself we can backport if needed.

PR #486, opened 2022-05-30
Upgrade dependencies

ThisIsMissEm:
Whilst working on some Inrupt code, I noticed that the jsonld.js dependencies were quite out of date, and this could be causing an issue we're seeing where in node, the http-client fails to load correctly.

This upgrades the direct dependencies to their latest versions, and fixes a minor issue that introduced regarding the 3xx Location redirects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ node_modules
 npm-debug.log
 tests/webidl/*-new
 v8.log
+test-suites
--- a/lib/documentLoaders/node.js
+++ b/lib/documentLoaders/node.js
@@ -143,7 +143,12 @@ module.exports = ({
           });
       }
       redirects.push(url);
-      return loadDocument(location, redirects);
+
+      // Ensure the redirect URL is resolved relative to the initial URL
+      // This supports relative Location headers and absolute:
+      const nextLocation = new URL(location, url).toString();
+
+      return loadDocument(nextLocation, redirects);
     }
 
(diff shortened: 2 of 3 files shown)


davidlehn:
I think pending changes that are in the v6.x branch right now may handle the major parts of this PR?  They came from here:
https://github.com/digitalbazaar/jsonld.js/pull/484

Any chance you could check if that branch works for your use cases?

It updates http-client and fixes a redirect issue from the test suite  It would be good to test that redirect code on more use cases too.  It didn't update other deps like lru-cache.  I had intended to wait on that and just handle the http-client issue that is causing the most trouble these days.

That v6.x branch unfortunately hasn't been merged into main and released yet because I didn't know how best to do so!  We like to keep linear history in main, but code got into main after 5.2.x and fixing it turned into a rather [large project](https://github.com/digitalbazaar/jsonld.js/pull/481) that still is not complete.  So it's unclear if we should do some evil rebasing in main or release on a branch and merge results back with a graph vs linear history.  I'll be figuring that out real soon now.

ThisIsMissEm:
The issue we see in http-client is that when the request happens, `httpClient.get` is undefined; looks like some sort of dependency injection doesn't work as expected. I didn't actually realise that there was a "next branch", but for now, we've worked around this by supplying a custom loader that is fetch based (it only needs to work for a specific use-case for now) — https://github.com/inrupt/solid-client-js/pull/1625

Issue #466, opened 2022-01-23
`compact` does not recognize string default when @type is not provided


jaxoncreed:
In the following example:

```typescript
  const inputData: JsonLd = [
    {
      "@id": "http://a.example/Employee7",
      "http://xmlns.com/foaf/0.1/givenName": [
        {
          "@value": "Robert",
        },
        {
          "@value": "Taylor",
        },
      ],
      "http://xmlns.com/foaf/0.1/familyName": [
        {
          "@value": "Johnson",
        },
      ],
      "http://xmlns.com/foaf/0.1/mbox": [
        {
          "@id": "mailto:rtj@example.com",
        },
      ],
    },
  ];
  const inputContext: ContextDefinition = {
    givenName: {
      "@id": "http://xmlns.com/foaf/0.1/givenName",
      "@type": "http://www.w3.org/2001/XMLSchema#string",
      "@container": "@set",
    },
    familyName: {
      "@id": "http://xmlns.com/foaf/0.1/familyName",
      "@type": "http://www.w3.org/2001/XMLSchema#string",
    },
    phone: { "@id": "http://xmlns.com/foaf/0.1/phone", "@container": "@set" },
    mbox: { "@id": "http://xmlns.com/foaf/0.1/mbox" },
  };
  const compactJsonLd = await compact(inputData, inputContext);
  console.log(compactJsonLd);

  // Logs:
  // {
  //     '@id': 'http://a.example/Employee7',
  //     'http://xmlns.com/foaf/0.1/familyName': 'Johnson',
  //     'http://xmlns.com/foaf/0.1/givenName': [ 'Robert', 'Taylor' ],
  //     mbox: { '@id': 'mailto:rtj@example.com' }
  //   }
```

Notice that "familyName" and "givenName" are not transformed. This is probably because these fields both of the `@type` `http://www.w3.org/2001/XMLSchema#string`. Setting the `@type` in the input data does not cause this problem. But, you shouldn't be required to provide a `@type` for strings as `http://www.w3.org/2001/XMLSchema#string` should be assumed if no type has been provided.

ethieblin:
I have the same issue when trying to round-trip from a framed JSON-LD to RDF and from the produced RDF to a framed JSON-LD.
```json
{
  "@context": {
    "td": "http://example.org/",
    "xsd": "http://www.w3.org/2001/XMLSchema#",
    "title": {
      "@id": "td:title",
      "@type": "xsd:string"
    },
[...]

PR #487, opened 2022-06-02
feat(compat): take into account native xsd types in "@type" in context

ethieblin:
fix https://github.com/digitalbazaar/jsonld.js/issues/466


davidlehn:
See comment in #466.  Spec clarification on the proper behavior and tests in the main test suites are needed.

PR #492, opened 2022-08-19
Use safe mode and `null` base by default in `canonize`.

dlongley:


davidlehn (approved):
Approved.  Please merge the basic tests first: https://github.com/digitalbazaar/jsonld.js/pull/493.

PR #497, opened 2022-08-29
Add expand and toRDF for non-IRI test.

davidlehn:
Test showing a potentially non-intuitive issue where `expand` passes safe mode but `toRDF` does not.


davidlehn:
I'll commit this since it's what the code does now.  But it might need some discussion later to determine if the expand behavior is what is best.

PR #503, opened 2022-11-09
Correct `canonize` documentation

dpetran:
The docs on the README for `canonize` state:

```js
const canonized = await jsonld.canonize(doc, {
  algorithm: 'URDNA2015',
  format: 'application/n-quads'
});
```

In order to properly accept n-quads the option needs to be:

```js
const canonized = await jsonld.canonize(doc, {
  algorithm: 'URDNA2015',
  inputFormat: 'application/n-quads'
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,7 +266,7 @@ const framed = await jsonld.frame(doc, frame);
 // (URDNA2015), see:
 const canonized = await jsonld.canonize(doc, {
   algorithm: 'URDNA2015',
-  format: 'application/n-quads'
+  inputFormat: 'application/n-quads'
 });
 // canonized is a string that is a canonical representation of the document
 // that can be used for hashing, comparison, etc.


davidlehn:
Something is a bit off here.  I think the intent in those docs is that `format` was for the output format, and the input `doc` is assumed to be JSON-LD.  With the `inputFormat` change here, that would run an N-Quads parser on the input instead of running `toRdf`.  Is that what you were trying to do?  Maybe just need to be clearer in examples that JSON-LD is the assumed input?

Looking here and in `rdf-canonize`, I'm not sure if `format` is even used currently?  It may be left over from legacy code, or was put in with the intent to allow other formats than N-Quads.  Due to how `rdf-canonize` is currently used, I'm not sure what other output would make sense.  But if we're accepting `format`, is should probably at least test that it is the supported `application/n-quads` value.

dpetran:
Ah, I was passing it an rdf document so that's how I assumed it was intended to be called (also, that's what the spec says the algorithm takes). I didn't continue reading the source past the `inputFormat`check, so I didn't think I could pass json-ld directly. 

Does output format make sense for this function? I thought `canonize` always returned a string of nquads.

gkellogg:
I think [rdflib.js](https://linkeddata.github.io/rdflib.js/doc/) uses a plugin architecture for parsers, and should be able to select an appropriate parser based on the MIME type. If you're going to do something with format, it would probably be good to not bake in too many formats, and maybe better leverage rdflib.js for finding parsers. Ideally, it should play well if the input is Turtle, TriG, or even RDFa, given loaded libraries that support it.

PR #507, opened 2023-02-02
Faster isObject

dlongley:


dlongley:
Added some other utility micro-optimizations that pay off with large datasets.

Issue #512, opened 2023-03-13
IPFS support in default document loader


F-Node-Karlsruhe:
We are currently looking into using IPFS as a context storage as it increases availability and conveniently takes care of versioning out of the box ("naming this with hashes"). We are using the `ipfs://CONTEXT_HASH` schema to reference the context.

The resolution of IPFS documents is currently implemented in the overwriting document loader [like so](https://github.com/european-epc-competence-center/vc-verifier/blob/d03a89ddc91e98a4926001cd7eb483e195fd46b1/api/src/services/fetch/index.ts#L48)

Is this something you would like to add in the future? If so i would make a PR regarding this topic.

The ipfs:// schema could be added here in the default document resolver
https://github.com/digitalbazaar/jsonld.js/blob/main/lib/documentLoaders/node.js#L59

F-Node-Karlsruhe:
Evolved into PR https://github.com/digitalbazaar/jsonld.js/pull/513

PR #513, opened 2023-03-21
init ipfs in default document loader

F-Node-Karlsruhe:
Simple IPFS support in the default document loader with origin https://github.com/digitalbazaar/jsonld.js/issues/512

Could be improved with the use of multiple public gateways and `Promise.any()` as shown in the issue above.

### DIFF
--- a/lib/documentLoaders/node.js
+++ b/lib/documentLoaders/node.js
@@ -56,13 +56,14 @@ module.exports = ({
   async function loadDocument(url, redirects) {
     const isHttp = url.startsWith('http:');
     const isHttps = url.startsWith('https:');
-    if(!isHttp && !isHttps) {
+    const isIpfs = url.startsWith('ipfs:');
+    if(!isHttp && !isHttps && !isIpfs) {
       throw new JsonLdError(
-        'URL could not be dereferenced; only "http" and "https" URLs are ' +
+        'URL could not be dereferenced; only "http", "https" and "ipfs" URLs are ' +
         'supported.',
         'jsonld.InvalidUrl', {code: 'loading document failed', url});
     }
-    if(secure && !isHttps) {
+    if(secure && isHttp) {
       throw new JsonLdError(
         'URL could not be dereferenced; secure mode is enabled and ' +
         'the URL\'s scheme is not "https".',
@@ -76,8 +77,10 @@ module.exports = ({
 
     let alternate = null;
 
+    const requestUrl = !isIpfs ? url : 'https://ipfs.io/ipfs/' + url.split('ipfs://')[1];
+
     const {res, body} = await _fetch({
-      url, headers, strictSSL, httpAgent, httpsAgent
+      requestUrl, headers, strictSSL, httpAgent, httpsAgent
     });
     doc = {contextUrl: null, documentUrl: url, document: body || null};
 
--- a/lib/documentLoaders/xhr.js
+++ b/lib/documentLoaders/xhr.js
@@ -32,22 +32,24 @@ module.exports = ({
   return queue.wrapLoader(loader);
 
   async function loader(url) {
-    if(url.indexOf('http:') !== 0 && url.indexOf('https:') !== 0) {
+    if(url.indexOf('http:') !== 0 && url.indexOf('https:') !== 0 && url.indexOf('ipfs:') !== 0) {
       throw new JsonLdError(
-        'URL could not be dereferenced; only "http" and "https" URLs are ' +
+        'URL could not be dereferenced; only "http", "https" and "ipfs" URLs are ' +
         'supported.',
         'jsonld.InvalidUrl', {code: 'loading document failed', url});
     }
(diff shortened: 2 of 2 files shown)


F-Node-Karlsruhe:
Example with W3C credential context:

`ipfs://QmcpYBo2zpfSKdBAqY8HPSoBhEoVAb1WjPP2muehvYcCgs` -> https://ipfs.io/ipfs/QmcpYBo2zpfSKdBAqY8HPSoBhEoVAb1WjPP2muehvYcCgs

dlongley (commented):
Hmm -- so at first glance at the PR title my thought was "We want to support IPFS but not in core." But looking at this PR, there's really not a whole lot here to make it happen.

@davidlehn what are your thoughts?

lib/documentLoaders/node.js
```diff
@@ -56,13 +56,14 @@ module.exports = ({
   async function loadDocument(url, redirects) {
     const isHttp = url.startsWith('http:');
     const isHttps = url.startsWith('https:');
-    if(!isHttp && !isHttps) {
+    const isIpfs = url.startsWith('ipfs:');
+    if(!isHttp && !isHttps && !isIpfs) {
       throw new JsonLdError(
-        'URL could not be dereferenced; only "http" and "https" URLs are ' +
+        'URL could not be dereferenced; only "http", "https" and "ipfs" URLs are ' +
```
dlongley, line 62:
```suggestion
        'URL could not be dereferenced; only "http", "https", and "ipfs" URLs are ' +
```

lib/documentLoaders/xhr.js
```diff
@@ -32,22 +32,24 @@ module.exports = ({
   return queue.wrapLoader(loader);
 
   async function loader(url) {
-    if(url.indexOf('http:') !== 0 && url.indexOf('https:') !== 0) {
+    if(url.indexOf('http:') !== 0 && url.indexOf('https:') !== 0 && url.indexOf('ipfs:') !== 0) {
       throw new JsonLdError(
-        'URL could not be dereferenced; only "http" and "https" URLs are ' +
+        'URL could not be dereferenced; only "http", "https" and "ipfs" URLs are ' +
```
dlongley, line 37:
```suggestion
        'URL could not be dereferenced; only "http", "https", and "ipfs" URLs are ' +
```

dlongley:
@davidlehn, also, if we do want to accept this PR, I recommend we should reconsider adding the URL itself (or at least the beginning of it if it's huge) to the error message string itself (I know it's already in the details) -- as that has been a common complaint since those details are often not surfaced cleanly in various tools.

F-Node-Karlsruhe:
> I think it would be nice if you could optionally provide your own IPFS host, with the current value as default.

That would be great. In my opinion an [semi-custom array](https://github.com/european-epc-competence-center/vc-verifier/blob/main/api/src/services/fetch/index.ts#L8) along with `Promise.any()` like [so](https://github.com/european-epc-competence-center/vc-verifier/blob/main/api/src/services/fetch/index.ts#L52) would be desirable, but may be too much for the core/default document loader.

dlongley:
@davidlehn can we get your thoughts here? I think we would like this functionality, it's just a question of how it's done.

nklomp:
Imo for VCI it should be pretty clear if you look at the spec and want something like this. Start with the issuer url in the DID doc, and follow normal OID4VCI discovery from there.

You are making sure that the issuer and well-knowns are served from the same origin this way. Our lib will not work if that is not the case, and we certainly wouldn't be adding something where well-knowns potentially might be served from different origins.

ntn-x2:
Hello! Has there been any progress lately on this PR?
Checking in because of https://github.com/OwnYourData/didlint/issues/3, which results in our DID method (`did:kilt`), not being properly verified by the DIDLint tool. We saw that replacing IPFS links with any HTTP-based gateway (e.g., `ipfs.io`), correctly verifies the document structure.

F-Node-Karlsruhe:
> Example with W3C credential context:
> 
> `ipfs://QmcpYBo2zpfSKdBAqY8HPSoBhEoVAb1WjPP2muehvYcCgs` -> https://ipfs.io/ipfs/QmcpYBo2zpfSKdBAqY8HPSoBhEoVAb1WjPP2muehvYcCgs

Based upon the [IPFS native url](https://docs.ipfs.tech/how-to/address-ipfs-on-web/#native-urls) scheme resolving towards an HTTP [IPFS gateway](https://docs.ipfs.tech/concepts/ipfs-gateway/#ipfs-gateway)

F-Node-Karlsruhe:
> I think it would be nice if you could optionally provide your own IPFS host, with the current value as default.

True that. The ipfs.io gateway has not been very reliable recently. One should carefully choose the default/s

https://cloudflare-ipfs.com seems to be a good choice as well

Btw. curl has released support for IPFS already https://curl.se/changes.html#8_4_0

Issue #523, opened 2023-06-01
expansion of `@id` doesn't work correctly if it contains a `:` and starts with a number


ndr-brt:
The expansion seems to be buggy on version 8.2.0.
If a `@id` value contains a ':' and it starts with a number it gets expanded in the wrong way

### How to replicate
From the node repl:
```
> const jsonld = require('jsonld');

> jsonld.expand({'@id': 'a3:42', '@type': 'Any'}).then(result => console.log(result))
> [ { '@id': 'a3:42', '@type': [ 'Any' ] } ] // CORRECT

> jsonld.expand({'@id': '3:42', '@type': 'Any'}).then(result => console.log(result))
> [ { '@id': '42', '@type': [ 'Any' ] } ] // WRONG
```

davidlehn:
Thanks. Addressed in https://github.com/digitalbazaar/jsonld.js/pull/533.

PR #527, opened 2023-08-01
fix: incorrect @id expansion

ndr-brt:
### What
This PR provides a way to avoid `@id` modification in case it contains `:` and it's not really an URI.
I'm not too familiar with the codebase so I just added a test case and a quick fix, but I'm more than available to adjust it based on feedback. 

### Why
We're trying to expand json documents that have `@id`s that contain `:` but they aren't really URIs.
The behavior has been described in #523.

Closes: #523

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,6 @@ coverage
 dist
 node_modules
 npm-debug.log
+package-lock.json
 tests/webidl/*-new
 v8.log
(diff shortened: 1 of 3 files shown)


ndr-brt:
@davidlehn sorry for tagging you here, but there's a possibility to get the PR reviewed? I cannot add reviewers by myself.

davidlehn:
Thanks. Addressed a bit different in https://github.com/digitalbazaar/jsonld.js/pull/533.

PR #528, opened 2023-08-11
Add react-native section to package.json

dmitrizagidulin:
Add React Native support, and instructions in the README.
See also https://github.com/digitalbazaar/rdf-canonize/pull/76


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # jsonld ChangeLog
 
+## 8.3.0 -
+### Added
```
dlongley, line 4:
```suggestion
## 8.3.0 -

### Added
```

README.md
```diff
@@ -364,6 +364,17 @@ The `safe` options flag set to `true` enables this behavior:
 const expanded = await jsonld.expand(data, {safe: true});
 ```
 
+### Using with React Native
+
+Using this library with React Native requires the
+[`data-integrity-rn`](https://github.com/digitalcredentials/data-integrity-rn)
```
dlongley, line 370: Seems an odd name for the polyfill. I wonder if there's a more general name like `polyfill-rn`? It's scoped to `@digitalcredentials` so may be that would be ok?

Hmm, maybe we also want to say that any polyfill will do, such as this one? That way we don't have a defacto dependency that we have to check with when doing updates.

dlongley (changes requested):

README.md
dmitrizagidulin: Rephrased to say 'a polyfill such as', which hopefully captures the same thought.

dmitrizagidulin:
@dlongley @davidlehn Any chance of merging in this PR, since it's got approval?

davidlehn:
- I made a different minimal patch here: https://github.com/digitalbazaar/jsonld.js/pull/572
- Text is aligned with what's in rdf-canonize.
- The types part left out since I don't know what is needed there, if anything, or how that interacts with other external types defs.
- The deps updates happened in another PR.

davidlehn:
Fixed in v9.0.0 via https://github.com/digitalbazaar/jsonld.js/pull/572.

PR #531, opened 2023-08-31
Fix graph property empty array and safe mode.

davidlehn:
- Fix handling of graph property with empty array.
- Fix safe mode for `@graph` use cases.
  - Check all elements of graph property with array.


dlongley (approved):
Thanks!

PR #539, opened 2023-10-02
Added URL to context resolution error message.

brianorwhatever:


lib/ContextResolver.js
```diff
@@ -171,8 +171,8 @@ module.exports = class ContextResolver {
       }
     } catch(e) {
       throw new JsonLdError(
-        'Dereferencing a URL did not result in a valid JSON-LD object. ' +
-        'Possible causes are an inaccessible URL perhaps due to ' +
+        'Dereferencing the URL ' + url + ' did not result in a valid JSON-LD ' +
```
dlongley, line 174:
```suggestion
        `Dereferencing the URL "${url}" did not result in a valid JSON-LD ` +
```

lib/ContextResolver.js
```diff
@@ -184,8 +184,8 @@ module.exports = class ContextResolver {
     // ensure ctx is an object
     if(!_isObject(context)) {
       throw new JsonLdError(
-        'Dereferencing a URL did not result in a JSON object. The ' +
-        'response was valid JSON, but it was not a JSON object.',
+        'Dereferencing the URL ' + url + ' did not result in a JSON object. ' +
```
dlongley, line 187:
```suggestion
        `Dereferencing the URL "${url}" did not result in a JSON object. ` +
```

dlongley (approved):
See suggestions. Thanks!

davidlehn:
- It's unfortunate to duplicate like this, but perhaps helpful.  In theory the details and cause should still be looked at to possibly find a lower level issue.  I'm not sure how to cleanly indicate this.
- Would it be better to put the URL at the end?  Would make the common text easier to read before the potentially long variable content.
- The node doc loader outputs something similar for regular invalid URLs.  It doesn't handle garbage URLs very well and puts them directly in the error message without good sanitizing.  I think that problem will happen before URLs get here.

dmitrizagidulin:
@davidlehn do you mind merging this PR, and handling the question of URLs at the end in a separate issue? This is really useful diagnostics.

davidlehn:
Replaced with https://github.com/digitalbazaar/jsonld.js/pull/561.  Closing.

PR #544, opened 2023-12-06
Fix `@graph` `@container` term `@context` handling.

davidlehn:
When a term scoped context is nullified, it also nullifies the container information. This fix gets the `@container` value from the active context. In particular, this is needed for proper `@graph` container handling.

It is a bit difficult to reason about the correctness of this fix.  All test suite tests including the additional one related to this issue pass.  But more esoteric untested JSON-LD structures may not?

https://github.com/w3c/json-ld-api/pull/582
https://github.com/w3c/vc-data-model/issues/1373

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # jsonld ChangeLog
 
+## 8.3.2 - 2023-xx-xx
+
+### Fixed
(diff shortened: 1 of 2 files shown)


BigBlueHat (approved):
This looks correct to me. Perhaps we need to round up the weirdest context files we can find for future testing. 😄

dlongley (approved):

PR #555, opened 2024-06-19
Update README to use default export and destructure on default export.

aljones15:
Updates the README so the es6 import examples work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,11 +168,11 @@ jspm install npm:jsonld
 ```
 
 ``` js
-import * as jsonld from 'jsonld';
+import jsonld from 'jsonld';
 // or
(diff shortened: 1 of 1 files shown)


dlongley (commented):
@davidlehn, is this going to change right back in the imminent next release? If so, let's not make this change, otherwise, +1.

BigBlueHat:
@davidlehn the CommonJS lines of code in this PR are wrong, but the need is real since `import * as jsonld from 'jsonld';` fails. I'd like to avoid more rocks in shoes.

I know you have preferences about no CJS stuff in the readme, but would you support a fix for these lines if I keep this using ESM?

PR #576, opened 2025-11-21
Update README with ES6 import example.

BigBlueHat:
Add ES6 import syntax for `jsonld` in README.

The other import styles will not work with Node.js, so this marks out the happy path for Node devs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,10 @@ npm install jsonld
 const jsonld = require('jsonld');
 ```
 
+```js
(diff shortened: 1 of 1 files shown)


davidlehn:
- Related: https://github.com/digitalbazaar/jsonld.js/pull/555.
- The ESM/CJS style mixing can get tricky sometimes.  I was hoping to avoid mentioning it at all and move to ESM where we can control it better.

dlongley (approved):

